PR #695, opened 2021-07-06
Update Kythera Setup Guide

fabioanderegg:
Biggest change is fixing the URL to the download, which is kind of important.

Also notes that it does not work with SDK builds right now, still working on that.
Some other minor updates regarding renaming of the Editor project.


AMZN-daimini (approved):

fabioanderegg:
We fixed the URL on our side, so not urgent any more.

fabioanderegg:
Tutorial is not ready yet, so remove link for now.

micronAMZN (approved):

PR #747, opened 2021-07-19
OS buttons in download page created

sptramer:
Signed-off-by: Kommineni <skommine@amazon.com>
Signed-off-by: Sindhu <33236959+sindhuk06@users.noreply.github.com>

changes as per PR comments

fixed code review comments

add active class to immidiate parent instead of all parent elements

OS buttons in download page created

download page os data view modification

hero imago copyright added

docs page navbar modifications

PR #757, opened 2021-07-21
OS buttons in download page created

AMZN-daimini:
NOTE - This change isn't meant to be submitted until an installer is made available on the O3DE.org website.

Restored OS buttons in Download page from this PR:
https://github.com/o3de/o3de.org/pull/747

Added minor fixes (changed default button to "Windows", fixed availability information)

image

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -80,6 +80,29 @@ $(function() {
   })
 });
 
+// download page os buttons
+
+$(function() {
+  const contentAvailable = "Fork the O3DE source and build the engine and tools directly from any branch. For more details, read the documentation.";
+  const contentUnavailable = "Coming soon.";
+  $(".download-os-button").click(function(event){
+    event.preventDefault()
+    
+    $(".download-os").removeClass("active");
+    $(this).parent().addClass("active");
+
+    if($(this).text().includes("Windows")){
+      $(".download-os-content").text(contentAvailable);
+      $("#download-page-buttons a").removeClass("disabled");
+      $("#download-page-buttons a").attr('href', 'https://github.com/o3de/o3de');
+    } else {
+      $(".download-os-content").text(contentUnavailable);
+      $("#download-page-buttons a").addClass("disabled");
+      $("#download-page-buttons a").removeAttr('href');
+    }
+  })
+});
+
 
 // Search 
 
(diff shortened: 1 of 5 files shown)


amzn-leenguy (commented):
I can sign off on the OS buttons, but the disabled CTA buttons for Mac OS and Linux, should we have it say Coming Soon (on the disabled buttons)? Or just leaved it as a disabled button with the original text? 

My preference is have them say Coming Soon.

layouts/partials/download/download.html
```diff
@@ -2,8 +2,8 @@ <h3>Get O3DE [Developer Preview]</h3>
 <div class="download_release">Release 2107.1</div>
```
sptramer, line 2: Missing a class for the "system requirements" to also be toggled along with the download button and text.

assets/js/app.js
```diff
@@ -80,6 +80,29 @@ $(function() {
   })
 });
 
+// download page os buttons
+
+$(function() {
```
sptramer, line 85: This function also needs to toggle the system requirements text.

assets/js/app.js
```diff
@@ -80,6 +80,29 @@ $(function() {
...
+    $(".download-os").removeClass("active");
+    $(this).parent().addClass("active");
+
+    if($(this).text().includes("Windows")){
+      $(".download-os-content").text(contentAvailable);
+      $("#download-page-buttons a").removeClass("disabled");
+      $("#download-page-buttons a").attr('href', 'https://github.com/o3de/o3de');
+    } else {
+      $(".download-os-content").text(contentUnavailable);
+      $("#download-page-buttons a").addClass("disabled");
```
sptramer, line 100: I don't think we should say that you can't or shouldn't fork from GitHub - there is "experimental" support for macOS and "in-progress" (maybe even considered as stable as Windows?) for Linux. Even if there's not an installer available users should be encouraged to fork.

assets/js/app.js
```diff
@@ -80,6 +80,29 @@ $(function() {
...
+  const contentAvailable = "Fork the O3DE source and build the engine and tools directly from any branch. For more details, read the documentation.";
+  const contentUnavailable = "Coming soon.";
+  $(".download-os-button").click(function(event){
+    event.preventDefault()
+    
+    $(".download-os").removeClass("active");
+    $(this).parent().addClass("active");
+
+    if($(this).text().includes("Windows")){
+      $(".download-os-content").text(contentAvailable);
```
sptramer, line 95: Windows will also have a binary installer available as part of this update - there should be a button for both "Fork from GitHub" and one for "Download Installer"

AMZN-daimini:
This is being reworked here: https://github.com/o3de/o3de.org/pull/1173

PR #774, opened 2021-08-03
Change gem/framework names to lowercase in API reference paths.

sptramer:
Signed-off-by: Stephen Tramer <stramer@amazon.com>


sptramer:
Merging to fix some issues with the broken link checker(s) caused by a mismatch between case names in links and the API reference path.

PR #791, opened 2021-08-06
fix a lot of broken links in physx docs

greerdv:
Signed-off-by: greerdv <greerdv@amazon.com>


moraaar (approved):
Thanks for fixing these.

I just noticed that if the links have changed that means probably the urls set in the engine for the help icon in the components are wrong now and have to be fixed. Could you double check the `AZ::Edit::Attributes::HelpPageURL` are still valid?

greerdv:
> Thanks for fixing these.
> 
> I just noticed that if the links have changed that means probably the urls set in the engine for the help icon in the components are wrong now and have to be fixed. Could you double check the `AZ::Edit::Attributes::HelpPageURL` are still valid?

I only fixed links from the docs to other parts of the docs, I didn't change any of the destination urls, so it shouldn't have any effect on the help page urls but I will double check.

sptramer (approved):
Broken link checker tool reports no more broken links to physx as a result of this PR.

greerdv:
> Broken link checker tool reports no more broken links to physx as a result of this PR.

Great, glad I got them all :) 

One question - does the broken link checker report broken links to other sections of the same document? e.g. like this?

[Another section of the same document](#not-the-right-link) 

## Another section

sptramer:
I don't believe the tool that I'm using currently does that, but I might be submitting PRs to it as part of figuring out how to integrate it into our approval check workflow.

greerdv:
> I don't believe the tool that I'm using currently does that, but I might be submitting PRs to it as part of figuring out how to integrate it into our approval check workflow.

Cool, thanks

PR #798, opened 2021-08-10
Remove preview banners, move shortcodes below intro to improve search…

micronAMZN:
… summaries, set zero length topics to draft. Atom Guide only.

Signed-off-by: Mike Cronin <mikecro@amazon.com>

### DIFF
--- a/content/docs/atom-guide/dev-guide/materials/material-build-pipeline.md
+++ b/content/docs/atom-guide/dev-guide/materials/material-build-pipeline.md
@@ -1,5 +1,6 @@
 ---
 title: Material Build Pipeline
+draft: true
 ---
 
 {{< placeholder >}}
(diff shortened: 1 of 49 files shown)


content/docs/atom-guide/dev-guide/materials/material-build-pipeline.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Material Build Pipeline
+draft: true
```
willihay, line 3: I ran a search and found 3 links into this doc. To avoid broken links, you'll want to disable or remove those.

willihay (changes requested):
To avoid creating new broken links, search for links into each of the topics you're setting to `draft` and disable or remove them. In some cases, such as with links from the Atom glossary, I think disabling the link would be the best move, to preserve the thinking that went behind creating those glossary links for when the placeholder topic is finally written.

willihay (approved):

PR #817, opened 2021-08-16
Update system requirements. 

chanmosq:
Added note about macOS support and Shader Model versions.

Signed-off-by: Chanelle Mosquera <chanmosq@amazon.com>


andreasplesch:
Had to look up what 'having' Shader Model 6.3 means. Seems to be DirectX associated and provided with gfx drivers at a certain hardware level. Perhaps most of the docs target audience would know but a link to a wiki entry or similar may be helpful.

content/docs/user-guide/components/reference/atom/diffuse-gi.md
```diff
@@ -7,6 +7,10 @@ toc: true
 
 The **Diffuse Global Illumination** component is a Level component that controls the quality level of global illumination that **Diffuse Probe Grid** components provide. The diffuse global illumination technique calculates the effect of light bounce, reflection, scatter, and absorption throughout an environment. It simulates real-world lighting behavior, where objects are lit both from the light source and from the light reflected by other objects.
 
+{{< note >}}
+You must have Shader Model 6.3 or later to use raytracing in the **Atom Renderer**.
```
willihay, line 11: As another reviewer commented, perhaps "having Shader Model 6.3" could be reworded to better define what we mean. Would it be more accurate, and sufficient, to say "Your GPU must support Shader Model 6.3" to use raytracing? If so, consider linking to [https://en.wikipedia.org/wiki/High-Level_Shading_Language](https://en.wikipedia.org/wiki/High-Level_Shading_Language). Or do we also need to mention graphic driver version support?

content/docs/welcome-guide/requirements.md
```diff
@@ -12,20 +12,26 @@ Open 3D Engine (O3DE) has a minimum set of hardware and software requirements fo
...
 + 40 - 100 GB of free disk space
 + 1366 x 768 px screen resolution
++ Shader Model 6.2 (or Shader Model 6.3 to use raytracing features)
 
 {{< note >}}
 RAM and free disk space requirements are dependent on the options that you select when configuring your project in O3DE.
 {{< /note >}}
 
+{{< note >}}
+Developing with O3DE on macOS is an experimental feature. At a minimum, you need macOS Catalina and XCode 11.
```
willihay, line 32: I think a note on OS support belongs in the next section on Software prereqs.

willihay (changes requested):

content/docs/user-guide/components/reference/atom/diffuse-gi.md
sptramer:
```suggestion
You must have a GPU with support for DirectX 12 Shader Model 6.3 or later to use raytracing in the **Atom Renderer**.
```

content/docs/welcome-guide/requirements.md
sptramer: Metal should also be pulled down into this section, along with the other macOS requirements - M1/ARM arch isn't supported in particular.

sptramer (changes requested):

content/docs/welcome-guide/requirements.md
chanmosq: What is the proper name for M1/ARM arch? I found it referred to as an "Apple R1 processor" that is "ARM-based". So I was planning to just refer to is as "Apple R1 processor" unless there is a more accurate title.

Also, do you suggest I keep this informatin in a note? Currently the docs are Windows (and now Linux) focused. If we were to add Macs to the mix, I think we need to further rearrange the list of requirements so it's clear. 

Since Mac support is still experimental and we don't have futher documentation on it, I lean towards keeping it in a note for now.

content/docs/user-guide/components/reference/atom/diffuse-gi.md
chanmosq: I'll change the wording to the one Stephen suggested. 

I'm not sure if we need to link to that wiki, since it doesn't exactly point to the Shader Model 6.3. It's moreso deeper information that customers can search themselves if they are interested in what Shader Models are. But for the most part, they just need to be aware that their GPU supports Shader Model 6.3.

chanmosq:
> 
> 
> Had to look up what 'having' Shader Model 6.3 means. Seems to be DirectX associated and provided with gfx drivers at a certain hardware level. Perhaps most of the docs target audience would know but a link to a wiki entry or similar may be helpful.

@andreasplesch I updated the docs to read "You must have a GPU with support for DirectX 12 Shader Model 6.3 or later to use raytracing in the Atom Renderer." I believe this is specific enough such that users can find more information on it if they need. (I left a comment about it here too, see https://github.com/o3de/o3de.org/pull/817#discussion_r690809714). 

Does this update make it more clear, or do you suppose a link to the wiki is necessary for clarity?

willihay (approved):

content/docs/welcome-guide/requirements.md
sptramer: I agree that mac support should be in the note. As far as chipset support: Probably the easiest way to say it is that "Intel x86_64" macs are supported.

sptramer (approved):

Issue #101, opened 2021-03-17
[Missing information] "C++ API reference" subchapter is not included  in "O3DE UI reflected property editor" chapter


LB-RobertBaczkowski:
Link to page : https://o3deorg.netlify.app/docs/tools-ui/component-library/uidev-reflected-property-editor-component/

The "O3DE UI reflected property editor" chapter is the only one in the "UI component library" which does not include a section on C++ API reference.

FiniteStateGit:
Fixed with PR https://github.com/o3de/o3de.org/pull/850 .

LB-RobertBaczkowski:
The page has not been updated yet, will close as soon as it happens.

PR #850, opened 2021-08-30
Add API ref links and Related links to card component. [Issue#101]

FiniteStateGit:
Added API ref links and Related links to the card component.
Fixes https://github.com/o3de/o3de.org/issues/101

Signed-off-by: FiniteStateGit 
<31811688+FiniteStateGit@users.noreply.github.com>


spotz (approved):
This looks good. Thanks!

amzn-leenguy (approved):
looks fine here

willihay (approved):
I see the value of the Related link here to the Card component, but I'm not sure of the value of the C++ API reference links to it. Maybe there's value in the convenience, though - since ReflectedPropertyEditor kind of goes hand in hand with a Card? I don't see any harm in it so I'm on board. Thanks for the contribution!

PR #873, opened 2021-09-04
Add shortcodes to note, warning, caution and important callouts

FiniteStateGit:


micronAMZN (approved):

content/docs/user-guide/packaging/asset-bundler/list-operations.md
```diff
@@ -8,7 +8,8 @@ weight: 500
 Asset list comparisons are rules provided to the `AssetBundlerBatch.exe` tool to determine which files should be included or excluded from the final bundle asset list. The asset list files have the suffix `.assetlist` and contain a flat list of paths and names of asset files. 
 
 {{< note >}}
-While the comparison operations use terms from set theory, they are not exactly the same as the actual set operations. This is particularly true of the [delta comparison operation](#asset-bundler-list-operations-delta), which includes a file entry in both sets if one of those files has been modified in the second set. Additionally, file pattern matching is not a set operation.
+While the comparison operations use terms from set theory, they are not exactly the same as the actual set operations. This is particularly true of the [
+comparison operation](#asset-bundler-list-operations-delta), which includes a file entry in both sets if one of those files has been modified in the second set. Additionally, file pattern matching is not a set operation.
```
sptramer, line 12:
```suggestion
While the comparison operations use terms from set theory, they are not exactly the same as the actual set operations. This is particularly true of the [delta comparison operation](#asset-bundler-list-operations-delta), which includes a file entry in both sets if one of those files has been modified in the second set. Additionally, file pattern matching is not a set operation.
```

content/docs/user-guide/scripting/lua/debugging-tutorial.md
```diff
@@ -48,7 +48,7 @@ This tutorial shows you how to use O3DE Lua Editor to perform debugging operatio
     end
 
     return ConstantRotation
-    ```
+    ```F
```
sptramer, line 51:
```suggestion
    ```
```

sptramer (changes requested):
Some minor fixes that look like bad automation or a mistype.

FiniteStateGit:
Thanks for catching those mistypes.

PR #912, opened 2021-09-15
Updated decal.md

mrieggeramzn:
Adding initial decal description.
Signed-off-by: mrieggeramzn mriegger@amazon.com


micronAMZN (changes requested):
Before we move forward we need to amend this commit with a sign-off so the DCO check can pass.

PR #1025, opened 2021-10-08
Add a banner for O3DEcon to Home, Download and Community pages

AMZN-daimini:
Adding a banner for O3DEcon to the website homepage

image
image

image


sptramer (approved):
Suggest a followup PR to add to "Community" and "Download" pages, but won't block.

AMZN-daimini:
> Suggest a followup PR to add to "Community" and "Download" pages, but won't block.

I think it makes more sense to move this to its own partial if we want it to be used in multiple pages. I can do that here directly.

PR #1037, opened 2021-10-13
Enable robots.txt generation

AMZN-alexpete:
This time with 100% more robots.txt per https://gohugo.io/templates/robots/


sptramer (approved):

sptramer:
Merging on admin authority since the lack of site indexing is actually a real, serious issue.

PR #1040, opened 2021-10-13
add detail about how infinite mass bodies behave

greerdv:
Signed-off-by: greerdv <greerdv@amazon.com>


greerdv:
Added to provide a better explanation for the case in https://github.com/o3de/o3de/issues/2767. I will also update the tooltip on the mass in a separate PR.

moraaar (approved):

amzn-sean (approved):
LGTM

PR #1048, opened 2021-10-15
Removing the term Perspective in favor of the general term viewport

micronAMZN:
because the pane is no longer named Perspective in the interface.

Signed-off-by: Mike Cronin <mikecro@amazon.com>


FiniteStateGit (changes requested):
Sorry I don't see a way to add line edits.

- The last bullet point of 'General terminology guidance' refers to Perspective.
- Should **Console** be **Editor Console** per the description?

PR #1049, opened 2021-10-15
Make search element id's unique.

FiniteStateGit:
Testing fix for broken search functionality in section navigation. Do not merge.


micronAMZN (commented):
Tested in deployment, left gutter search works, header search now seems borked.

PR #1060, opened 2021-10-18
Replace current O3DEcon banner with an O3DEjam one.

AMZN-daimini:
Thanks to @amzn-leenguy for the svg logo :D

Desktop
image

Mobile
image


layouts/partials/top-banner.html
```diff
@@ -1,10 +1,10 @@
 <a class="top-banner" href="https://events.linuxfoundation.org/o3decon/" target="_blank">
```
micronAMZN, line 1: Link goes to O3DE con page, not Game Jam info.

PR #1087, opened 2021-10-22
[o3de.org] Fixed small typos and website update

RexC2I:
-Fixed some typos
-On the website, under the main 'Contribute' page,  changed SIG-Presentation  to the correct name,  SIG-Graphics-Audio

- [x]  Does the documentation show the user something *meaningful*?

### DIFF
--- a/layouts/partials/contribute/sigs.html
+++ b/layouts/partials/contribute/sigs.html
@@ -20,7 +20,7 @@
       SIG-Platform
     </a>
     <a href="https://github.com/o3de/sig-presentation" class="card sig-card">
-      SIG-Presentation
+      SIG-Graphics-Audio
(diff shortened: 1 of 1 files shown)


layouts/partials/contribute/sigs.html
```diff
@@ -20,7 +20,7 @@
       SIG-Platform
     </a>
     <a href="https://github.com/o3de/sig-presentation" class="card sig-card">
```
AMZN-daimini, line 22: GitHub seems to correctly redirect this url, but we may as well change it to the correct one while we're at it :D
https://github.com/o3de/sig-graphics-audio

AMZN-daimini (approved):

PR #1176, opened 2021-11-15
Website - Scroll to anchors with correct offset.

AMZN-daimini:
Add some javascript to detect whenever an anchor link is clicked, and ensure the correct position is calculated (previously, the page would move to the right spot, but the top navbar was not taken into account).

The page will now scroll to the position instead of teleporting to it.

website_scrollAnimateOffset

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -160,3 +160,18 @@ $(function() {
     }
   });
 });
+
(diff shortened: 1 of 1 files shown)


FiniteStateGit (approved):
Approved.

FiniteStateGit (approved):

sptramer (approved):

PR #1179, opened 2021-11-15
Fix links | Atom Guide > Passes

chanmosq:
## Change summary

Fix links, by replacing / with .md to reference a page. NOTE: The link works with / in o3de.org, but not in github.com

### Submission Checklist:

### DIFF
--- a/content/docs/atom-guide/dev-guide/passes/_index.md
+++ b/content/docs/atom-guide/dev-guide/passes/_index.md
@@ -23,6 +23,6 @@ Passes are a means to produce a desired visual effect. To demonstrate, consider 
 This section covers the following topics.
 | Topics                        | Description |
 |--------------------------------------|---------|
-| [Pass System](pass-system/) | An overview of the pass system in Atom. |
-| [Pass Template File Specification](pass-template-file-spec/) | JSON reference for pass template data (`*.pass` files). |
-| [Authoring Passes](authoring-passes/) | Learn how to author passes in the pass system.  |
+| [Pass System](pass-system.md) | An overview of the pass system in Atom. |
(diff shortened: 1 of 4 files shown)


micronAMZN (commented):
This is odd. In my prior experience, adding the `.md` extension to a link breaks the link. This seems to be working in preview, though. Just the file name or (directory path for index files) without the extension or slash should be sufficient for these links. Waiting on additional feedback before approval.

willihay:
In the description of this PR it sounds like the purpose is to allow readers to follow links in case they're browsing from github.com instead of the website. Is that correct?

Our style guide seems to preclude this. The guidance on the use of [links](https://o3de.org/docs/contributing/to-docs/style-guide/format/#links) states: "Write links without the file extension _(sic)_ .md."

Are you thinking this guidance should be changed to allow browsing from github.com? If so, perhaps this should be brought up in a sig-docs meeting to discuss a change of guidance followed by a mass update to all links. I'm not sure it makes sense to change just a few links on our repo. That inconsistency could be frustrating to readers.

willihay (commented):
Holding off on approval pending discussion of earlier comments.

chanmosq:
Thanks for the further investigation. Given the Style Guide's guidelines for formatting links, and the state of all O3DE documentation following these guidelines, I agree with @willihay that this would cause inconsistency in style across O3DE docs and is an unnecessary change.  

For context, the reason I created this PR is because some developers use [o3de/o3de.org repo](https://github.com/o3de/o3de.org) to navigate through the documentation. They expected to be able to click on the link and go to the page in the repository. This is a reasonable expectation, but we can defer making changes until there's more data indicating that this is a problem. If SIG Docs and Community later decide to make that change, we should update the Style Guide and carefully change all the links. 

Closing this PR.

PR #1180, opened 2021-11-16
Website - Add slideshow for homepage hero section.

AMZN-daimini:
Introduces an animated slideshow for the homepage hero section.
Currently, we're still aiming to have a video looping in the background, but this is a backup solution in case we have issues procuring the content in time, or identifying where to store it.

Of course the images used are samples that were randomly picked from the doc pages, @amzn-leenguy is working to select the final ones. They may go in as a separate submission depending on timing.

home_hero_slideshow

(had to compress the gif hard but the actual result is smoother :) )


erickson-doug (approved):

yuyihsu (commented):
We will need to replace the still images with a few other options, which I will send out later. but overall I approve this pull request.

PR #1181, opened 2021-11-16
Update resource mapping file docs to reflect optional account ID

allisaurus:
Signed-off-by: Stanko <34254888+allisaurus@users.noreply.github.com>

## Change summary

Update docs to reflect schema changes in https://github.com/o3de/o3de/pull/5475

### Submission Checklist:

### Testing

Screenshot (updated 11/17/21) of `/user-guide/gems/reference/aws/aws-core/resource-mapping-files/`
image

### DIFF
--- a/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
+++ b/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
@@ -32,7 +32,7 @@ Each entry defines a lookup name - a logical name for the resource - and a colle
 
 The mapping file also supports one _optional_ and two **required** global attributes:
 
-* AccountId (optional) - The default AWS account that is used to search for AWS resources referenced in your O3DE application code.
+* AccountId (optional) - The default AWS account that is used to search for AWS resources referenced in your O3DE application code. Used when account ID is not otherwise explicitly referenced (such as on a specific resource mapping).
 * Region - The default region to use with this mapping file.
 * Version - The json schema version of this document.
 
--- a/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
+++ b/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
@@ -29,9 +29,10 @@ Each entry defines a lookup name - a logical name for the resource - and a colle
 * AccountId - Used to override the global account id attribute, if this resource is deployed in a different AWS account.
 
 **Global attributes**
-The mapping file also supports three **required** global attributes:
 
-* AccountId - The default AWS account containing resources in this mapping file.
+The mapping file also supports one _optional_ and two **required** global attributes:
+
+* AccountId (optional) - The default AWS account containing resources in this mapping file.
 * Region - The default region to use with this mapping file.
 * Version - The json schema version of this document.
 
@@ -52,6 +53,10 @@ The mapping file also supports three **required** global attributes:
 ```
 
 This mapping file has a single mapping: **MyKey**, which is a Lambda function called "ExampleLambda". It is deployed in us-west-2 in the AWS account with account ID "123456789123".
+
+{{< important >}}
+Your AWS account ID should not be included in builds running on untrusted clients.
+{{< /important >}}
 
 ## Generating a resource mappings file
 
--- a/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
+++ b/content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
@@ -32,7 +32,7 @@ Each entry defines a lookup name - a logical name for the resource - and a colle
 
 The mapping file also supports one _optional_ and two **required** global attributes:
 
-* AccountId (optional) - The default AWS account containing resources in this mapping file.
+* AccountId (optional) - The default AWS account that is used to search for AWS resources referenced in your O3DE application code.
 * Region - The default region to use with this mapping file.
 * Version - The json schema version of this document.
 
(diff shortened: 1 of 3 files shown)


content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
```diff
@@ -53,6 +54,10 @@ The mapping file also supports three **required** global attributes:
 
 This mapping file has a single mapping: **MyKey**, which is a Lambda function called "ExampleLambda". It is deployed in us-west-2 in the AWS account with account ID "123456789123".
 
+{{< important >}}
+Your AWS account ID should not be included in builds running on untrusted clients.
```
lmbr-pip, line 58: Builds seems odd in this context. Maybe, We strongly recommend not including AWS account id in resource mapping files in release or other distributed builds. Utilize access patterns like [APIGateway and Custom domain](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html) names to provide alias points to resources.

Not sure if theres a catch all blog we can use, so this may be tricky. But we need to provide help in what to do instead.

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
```diff
@@ -29,9 +29,10 @@ Each entry defines a lookup name - a logical name for the resource - and a colle
 * AccountId - Used to override the global account id attribute, if this resource is deployed in a different AWS account.
 
 **Global attributes**
-The mapping file also supports three **required** global attributes:
 
-* AccountId - The default AWS account containing resources in this mapping file.
+The mapping file also supports one _optional_ and two **required** global attributes:
+
+* AccountId (optional) - The default AWS account containing resources in this mapping file.
```
lmbr-pip, line 35: Not sure why region is required, but thats for another PR. Its just there as a helper metadata so you don't have to set it everywhere.

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-tool.md
```diff
@@ -170,4 +170,4 @@ Check the [resource mapping file](/docs/user-guide/gems/reference/aws/aws-core/r
 Errors of the form ```Cannot determine region from the url``` in the logs indicate the wrong region or account has been set for the resource.
 * Double check that the resource's mapping entry has the correct region and account information, if overriding the global region and account attributes.
 * Double check that the resource's mapping entry has the correct resource type and name.
-* Ensure the mapping file sets the required global attributes for region and account. Ensure they are set to the expected values.
+* Ensure the mapping file sets the required global attribute for region and that it is set to the expected value.
```
lmbr-pip, line 173: nit: technically this should be "global attributes for region and version"

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
junbo75: If I remember it correctly, AWSCore is using that region information to construct the API Gateway endpoint for service API calls.

junbo75: Maybe change it to something like: "The default AWS account that is used to search for AWS resources referenced in your O3DE application code"?

This global account ID may not be the account that contains the resource in the resource mapping file since each resource may has it own account ID.

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-tool.md
```diff
@@ -170,4 +170,4 @@ Check the [resource mapping file](/docs/user-guide/gems/reference/aws/aws-core/r
 Errors of the form ```Cannot determine region from the url``` in the logs indicate the wrong region or account has been set for the resource.
 * Double check that the resource's mapping entry has the correct region and account information, if overriding the global region and account attributes.
 * Double check that the resource's mapping entry has the correct resource type and name.
-* Ensure the mapping file sets the required global attributes for region and account. Ensure they are set to the expected values.
+* Ensure the mapping file sets the required global attributes for region and version. Ensure they are set to the expected values.
```
junbo75, line 173: Is the version decided the schema we use or do we allow developers to set it to be any valid value?

allisaurus: I think the intention is that `version` reflects what schema version the file reflects w/r/t what we support. However, we [currently allow this](https://github.com/o3de/o3de/blob/development/Gems/AWSCore/Code/Include/Private/ResourceMapping/AWSResourceMappingConstants.h#L66) to be any "X.X.X" like value, so I don't think there's anything stopping developers from setting this right now to something other an what we've used in the past.

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-files.md
```diff
@@ -29,9 +29,10 @@ Each entry defines a lookup name - a logical name for the resource - and a colle
 * AccountId - Used to override the global account id attribute, if this resource is deployed in a different AWS account.
 
 **Global attributes**
-The mapping file also supports three **required** global attributes:
 
-* AccountId - The default AWS account containing resources in this mapping file.
+The mapping file also supports one _optional_ and two **required** global attributes:
+
+* AccountId (optional) - The default AWS account that is used to search for AWS resources referenced in your O3DE application code.
```
lmbr-pip, line 35: nit: The default  aws account id to use when no account id is explicitly referenced ?

content/docs/user-guide/gems/reference/aws/aws-core/resource-mapping-tool.md
lmbr-pip: We do not enforce version, we just enforce version format. Devs can set what they want but we use it to to show the version are tooling is generating from. Version number is owned by published json schema, which I believe is only in docs currently.

lmbr-pip (approved):
LG, one minor change to report.

onecent1101 (approved):

allisaurus:
FYI: `do-not-merge/in-review` applied pending OK from @o3de/docs-reviewers

sptramer (approved):

PR #1183, opened 2021-11-16
Website - Refactor Getting Started section in homepage

AMZN-daimini:
Section now uses a carousel to prevent it from getting too long.
Carousel does not scroll automatically but can be operated by the user.

image

On tablet resolution, we revert to rows as the 3 columns were too narrow

image

In the mobile version, I removed the images since they were being cropped and unreadable

image


erickson-doug (approved):

erickson-doug (approved):

AMZN-daimini:
I fixed the sizing so all cards have the same height, the image has been updated accordingly.

@yuyihsu if it looks ok, we need another approval to submit this to the live website :)

yuyihsu (approved):

PR #1189, opened 2021-11-18
Website - Add Community Collaboration section to Homepage

AMZN-daimini:
Adding a Community Collaboration section to the homepage.

Note: the image from Deadhaus Sonata is intentionally not part of this submission. We will need it to be uploaded on Apocalypse's server and displayed from there.

collaboration


AMZN-daimini:
Sorry for the base switching, had to update staging to display the correct diffs :)

bhanuja-s (changes requested):
Brilliant, thank you for the GIF! A few comments:
- The button for the second card should be "View" as it would go to a new tab.
- We don't need the "View more projects" button for this release.
- There should be a 20% overlay (#000000) for the card, which should be 50% on the hover state.

erickson-doug (approved):

AMZN-daimini:
> Brilliant, thank you for the GIF! A few comments:
> 
> * The button for the second card should be "View" as it would go to a new tab.
> * We don't need the "View more projects" button for this release.
> * There should be a 20% overlay (#0) for the card, which should be 50% on the hover state.

Neat, I will apply these changes in a separate PR

PR #1220, opened 2021-11-20
Hide visualization/cameras and environments

### LINKED ISSUE #240: [Blank Pages]: O3DE User Guide
The following pages are completely empty and do not have neither placeholder information nor a preview window.

- [ ] https://o3deorg.netlify.app/docs/user-guide/visualization/cameras/
- [ ] https://o3deorg.netlify.app/docs/user-guide/visualization/cameras/setup-and-rigging/
- [ ] https://o3deorg.netlify.app/docs/user-guide/visualization/environments/
- [x] https://o3deorg.netlify.app/docs/user-guide/visualization/rendering/
- [x] https://o3deorg.netlify.app/docs/user-guide/visualization/rendering/atom-in-o3de/
- [x] https://o3deorg.netlify.app/docs/user-guide/visualization/rendering/shaders/

FiniteStateGit:
Hide empty section links for visualization/cameras and visualization/environments.

Fix #240


micronAMZN (approved):

content/docs/user-guide/gems/core-gems.md
```diff
@@ -16,7 +16,7 @@ These Gems are only built and loaded for products that require them:
 
 * **Atom** - The real-time physically based renderer used by O3DE. In addition to rendering at runtime, Atom provides Editor components and asset processing tools. For more information on Atom, read the [Atom Guide](/docs/atom-guide).
 
-* **CameraFramework** - Components for placing and controlling cameras. Read the [Cameras documentation](/docs/user-guide/visualization/cameras/) for more information.
+* **CameraFramework** - Components for placing and controlling cameras. Read the [Cameras documentation](/docs/user-guide/components/reference/camera/) for more information.
```
willihay, line 19: The link to reference/camera, while valid, does not resolve well [in practice](https://deploy-preview-1220--o3deorg.netlify.app/docs/user-guide/components/reference/camera/) because the _index is blank, only serving to establish the collapsible section.
```suggestion
* **CameraFramework** - Components for placing and controlling cameras. Read the [Cameras documentation](/docs/user-guide/components/reference/camera/camera) for more information.
```

willihay (changes requested):

content/docs/user-guide/gems/core-gems.md
FiniteStateGit: Ok thank you, I was thinking that page would exist to point to the Camera and Camera Rig components, if not now, at some point.  I will change this.  The Shape component's category *did* get a landing page, but it has its own interface to document.  Do you think that a landing page for component category would be an exception to a general rule?  Related, should the non-uniform scale component (or a singular component without other components in the same category) be a top level TOC item as the Transform component is?

willihay: Yes, it seems to be working out that way - that a component category having its own landing page, like the Shape component, is an exception. I think it just depends on the needs of the category. Although I'm not happy with how the website seems to be unable to deal with links to components/reference/camera, for example. The link resolves, but nothing is highlighted in the lefthand / ToC nav. Perhaps this could be fixed? Might be worth a GitHub issue.

willihay: And yes, I think the non-uniform scale component should be at the same level as the transform component, not by itself under its own collapsible section. Good catch!

willihay: The only exception to this example might be if more components were expected to appear under a category like that. But in the case of the non-uniform scale, I don't think this is the case.

willihay (approved):

Issue #1253, opened 2021-12-01
[ISSUE][DOCS] The command line to build from source in windows has an invalid parameter in it


lawsonamzn:
## Describe the issue

If you attempt to build from source code following the instruction on the below page, MSBuild will issue an error because the correct flag to pass is -m, instead of /m (ie, no slash, must be dash).  See https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019

## Impacted page(s)

https://github.com/o3de/o3de.org/blob/main/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
which turns into 
https://o3de.org/docs/welcome-guide/setup/setup-from-github/building-windows/

## Expected experience

The command line parameter should result in msbuild functioning.

lawsonamzn:
Actually, I cannot reproduce this.  It works with /m or -m for me.
Tried on
* Windows CMD
* windows CMDR
* windows powershell 5
* windows powershell 7
* All of the above in the new windows terminal

lawsonamzn:
fixed by PR #1254
We discussed this in Discord, and its just safer to recommend -m than /m.  There are some limited conditions (probably Cygwin, or other bash-like shells) which will interpret /m as the folder named /m, wheras they all accept -m as the switch -m.

PR #1254, opened 2021-12-01
Fixes MSBuild compile option (/m should be -m)

lawsonamzn:
Signed-off-by: lawsonamzn <70027408+lawsonamzn@users.noreply.github.com>

## Change summary
Fixes bug #1253 - The compile parameter for msbuild is actually -m, not /m, and /m is interpreted as a folder name, resulting in an error.
See https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019

### Submission Checklist:

## Impacted page(s)

https://github.com/o3de/o3de.org/blob/main/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
which turns into
https://o3de.org/docs/welcome-guide/setup/setup-from-github/building-windows/

### DIFF
--- a/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
+++ b/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
@@ -56,10 +56,10 @@ To prepare to build the engine and projects, choose one of the following build t
     The following example shows the `profile` build configuration.
 
     ```cmd
-    cmake --build build/windows_vs2019 --target Editor --config profile -- /m
+    cmake --build build/windows_vs2019 --target Editor --config profile -- -m
     ```
 
-    The `/m` is a recommended build tool optimization. It tells the Microsoft compiler (MSVC) to use multiple threads during compilation to speed up build times.
+    The `-m` is a recommended build tool optimization. It tells the Microsoft compiler (MSVC) to use multiple threads during compilation to speed up build times.
 
     The engine takes a while to build. If you've used all the example commands in these steps, when the build is complete, you can find the engine tools and other binaries in `C:\o3de\build\windows_vs2019\bin\profile`.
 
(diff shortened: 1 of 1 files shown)


burelc-amzn:
I don't think this is necessary. From that [linked documentation page](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019#:~:text=Every%20switch%20is%20available%20in%20two%20forms%3A%20%2Dswitch%20and%20/switch.%20The%20documentation%20only%20shows%20the%20%2Dswitch%20form.):
>Every switch is available in two forms: -switch and /switch. The documentation only shows the -switch form. 

So I believe `-m` and `/m` are synonymous.

lawsonamzn:
> I don't think this is necessary. From that [linked documentation page](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019#:~:text=Every%20switch%20is%20available%20in%20two%20forms%3A%20%2Dswitch%20and%20/switch.%20The%20documentation%20only%20shows%20the%20%2Dswitch%20form.):
> 
> > Every switch is available in two forms: -switch and /switch. The documentation only shows the -switch form.
> 
> So I believe `-m` and `/m` are synonymous.

I'll give it a test, but the user in question (in discord) literally could not get it to work until they changed /m to -m
It could be a CMAKE thing

CMAke might interpret / after -- differently than MSBuild.
\

willihay (approved):

lawsonamzn:
So in the discord it was discussed and decided we should just go with -m instead of /m since there are situations where /m might fail but -m is more compatible anyway.  Its likely its a weird shell like bash-for-windows or cygwin that is interpreting / as a directory instead of as a switch.

sptramer (approved):

PR #1328, opened 2021-12-31
Quick fix making more clear what --config means.

Jackerty:
I had an unpleasant compilation experience.

--config profile being the default was one of the causes as it turns on -Werror which then finds unused variable.
I won't say this is the best fix as
1) building-windows may have a similar issue
2) better explanation of what different configurations types do beyond the names would be better

My other problems are probably caused by myself. It was a weird thing where I had to run first cmake let it complain that it didn't find clang then run again with CMAKE_CXX_COMPILE. If CMAKE_CXX_COMPILE was in the first run cmake would complain about pthread.h missing. I think I fixed that with LY_UNITY_BUILD=OFF.

### DIFF
--- a/content/docs/welcome-guide/setup/setup-from-github/building-linux.md
+++ b/content/docs/welcome-guide/setup/setup-from-github/building-linux.md
@@ -60,6 +60,7 @@ To prepare to build the engine and projects, choose one of the following build t
     ```
 
     The `-j` is a recommended build tool optimization. It tells the Ninja build tool the number of parallel build tasks that will be executed simultaneously. The 'number of parallel build tasks' is recommended to match the number of cores available on the Linux host machine.
+    The `--config` sets the build configuration type: `profile`, `debug`, or `release`. For setting up O3DE, `profile` is recommended. Read more on [O3DE's build configurations](/docs/user-guide/build/configure-and-build.md#generated-build-configurations).
 
     Example:
 
@@ -107,6 +108,7 @@ To prepare to build the engine and projects, choose one of the following build t
     ```
 
     The `-j` is a recommended build tool optimization. It tells the Ninja build tool the number of parallel build tasks that will be executed simultaneously. We recommend that the 'number of parallel build tasks' matches the number of cores available on the Linux host machine.
+    The `--config` sets the build configuration type: `profile`, `debug`, or `release`. For setting up O3DE, `profile` is recommended. Read more on [O3DE's build configurations](/docs/user-guide/build/configure-and-build.md#generated-build-configurations).
 
     Example:
 
--- a/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
+++ b/content/docs/welcome-guide/setup/setup-from-github/building-windows.md
@@ -60,6 +60,7 @@ To prepare to build the engine and projects, choose one of the following build t
     ```
 
     The `-m` is a recommended build tool optimization. It tells the Microsoft compiler (MSVC) to use multiple threads during compilation to speed up build times.
+    The `--config` sets the build configuration type: `profile`, `debug`, or `release`. For setting up O3DE, `profile` is recommended. Read more on [O3DE's build configurations](/docs/user-guide/build/configure-and-build.md#generated-build-configurations).
 
     The engine takes a while to build. If you've used all the example commands in these steps, when the build is complete, you can find the engine tools and other binaries in `C:\o3de\build\windows_vs2019\bin\profile`.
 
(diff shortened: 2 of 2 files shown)


content/docs/welcome-guide/setup/setup-from-github/building-linux.md
```diff
@@ -103,10 +103,11 @@ To prepare to build the engine and projects, choose one of the following build t
 1. Use CMake to build the engine as an SDK, the same as if you installed the engine from an installer tool. The following example shows the `profile` build configuration.
```
AMZN-alexpete, line 103: this statement is no longer valid because the config now shows the 3 options.

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
```diff
@@ -103,10 +103,11 @@ To prepare to build the engine and projects, choose one of the following build t
...
 
     ```shell
-    cmake --build build/linux --target INSTALL --config profile -j <number of parallel build tasks>
```
AMZN-alexpete, line 106: For development profile is the recommended build type, debug is much slower to build and run and release is generally optimized for game releases.    If/when you build with the release config I'm not sure if tools like the editor will get built, what result did you get when you tested this?

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
Jackerty: You do have a point.

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
Jackerty: I didn't run --target INSTALL nor --config release so I don't know. I run with --target editor and --config debug.
I would just like to put a note that other build configurations exist and warning about "-Werror".

sptramer (approved):

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
```diff
@@ -103,10 +103,11 @@ To prepare to build the engine and projects, choose one of the following build t
...
+    cmake --build build/linux --target INSTALL --config <debug|profile|release> -j <number of parallel build tasks>
```
chanmosq, line 106: I think we should keep the example for `profile`. I understand the purpose is to make it clear that O3DE supports different configurations, but the configuration recommended for this workflow is `profile`, and as a "Getting Started" topic, that's most likely enough info for newcomers to know. 

I very much agree that knowing the different configurations are useful, and that information is difficult to find. So in my next comments, I make some suggestions to address that. 
```suggestion
    cmake --build build/linux --target INSTALL --config profile -j <number of parallel build tasks>
```

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
```diff
@@ -103,10 +103,11 @@ To prepare to build the engine and projects, choose one of the following build t
...
     ```
 
     The `-j` is a recommended build tool optimization. It tells the Ninja build tool the number of parallel build tasks that will be executed simultaneously. We recommend that the 'number of parallel build tasks' matches the number of cores available on the Linux host machine.
+    The `--config` sets configuration type. This affects compilation flags. For example, `profile` turns on `-Werror`.
```
chanmosq, line 110: Mention the different build configuration types and link to the docs where the three are clearly defined. 
```suggestion
    The `--config` sets the build configuration type: `profile`, `debug`, or `release`. For setting up O3DE, `profile` is recommended. Read more on [O3DE's build configurations](/docs/user-guide/build/configure-and-build.md#generated-build-configurations).
```

If you accept this suggestion, can you also put it in the **Source engine** tab?

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
```diff
@@ -103,10 +103,11 @@ To prepare to build the engine and projects, choose one of the following build t
...
 
     Example:
```
chanmosq, line 113: This comment is for line 119 (after "The engine takes a while to build..."), as I'm unable to make a comment there. 

In a "caution" callout box, we can mention the problem you encountered and a workaround. 
```suggestion

    {{< caution >}}
The `profile` configuration type enables the `-Werror` compilation flag. If you encounter build errors due to this, try building with the `debug` configuration instead. 
    {{< /caution >}}
    
```
----------------

Altogether, these suggested changes would render like this: 
image

chanmosq (changes requested):
@Jackerty I made a few suggestions to address your problem with `-Werror`, and finding info on the configuration types.

content/docs/welcome-guide/setup/setup-from-github/building-linux.md
Jackerty: I have learned that every profile enables -Werror so this is not correct information.

See:
https://github.com/o3de/o3de/blob/development/cmake/Platform/Common/Clang/Configurations_clang.cmake

The problem I had was caused by git somehow getting commit where debug variable was unused in other profiles.

Jackerty:
This push request is changed by adding a link to a location in the docs explaining build configurations.
I also added links to the windows build guide.

chanmosq (approved):
Looks good! Thank you for clearing up these instructions.

PR #1336, opened 2022-01-04
Move search results to dedicated page

AMZN-alexpete:
## Change summary

- moves search results to a dedicated page at `/search` 
- allows for hyperlinks that go directly to a search page e.g. `o3de.org/search?query_param=search+term`
- removes hidden giant search term blob from every page except the search page
- allows the back button to work and send the user to the previous page before the search was initiated

## Testing
I've tested this locally, but not on mobile.
One major change with this is the page header will look different when they go to the `/search` page because they're "leaving" whatever section of the site they're in e.g. "docs", "download" etc.  Those sub-sections have a blue header, but the search page will have the default black header you see on the main page.

Signed-off-by: Alex Peterson <26804013+AMZN-alexpete@users.noreply.github.com>

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -57,39 +57,12 @@ $(function() {
 
 $(() => {
   $("#search").submit(function(event){
-    event.preventDefault();
-    $(".search-input-view").show();
-
-    let data =  $('#search-input').val()
-    if (data.length > 0) {
-      const idx = lunr(function () {
-        this.ref('id')
-        this.field('title', {
-          boost: 15
-        })
-        this.field('tags')
-        this.field('content', {
-          boost: 10
-        })
-    
-        for (const key in window.store) {
-          this.add({
-            id: key,
-            title: window.store[key].title,
-            tags: window.store[key].category,
-            content: window.store[key].content
-          })
-        }
-      })
-    
-      // Perform the search
-      const results = idx.search(data)
-      // Update the search title
-      $('#search-title').text('Search Results for ' + '"' + data + '"')
-      // Update the list with results
-      displayResults(results, window.store)
+    if($(".search-input-view").is(":hidden"))
+    {
+      event.preventDefault();
+      $(".search-input-view").show();
     }
-  })
+  });
 });
 
 // --- Downloads page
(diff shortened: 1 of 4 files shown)


assets/js/app.js
```diff
@@ -57,39 +57,12 @@ $(function() {
...
-      // Perform the search
-      const results = idx.search(data)
-      // Update the search title
-      $('#search-title').text('Search Results for ' + '"' + data + '"')
-      // Update the list with results
-      displayResults(results, window.store)
+    if($(".search-input-view").is(":hidden"))
+    {
+      event.preventDefault();
+      $(".search-input-view").show();
```
AMZN-daimini, line 63: It would make sense for this to be a toggle, so that it can be closed again?

It can be a separate discussion with UX though.

AMZN-daimini (approved):
Looks good to me.

One thing I noticed, though, is that when going through the search flow the url also shows x and  y params
image
I don't think it's a big problem, but I'm curious what is injecting them there.

AMZN-alexpete:
> Looks good to me.
> 
> One thing I noticed, though, is that when going through the search flow the url also shows x and y params image I don't think it's a big problem, but I'm curious what is injecting them there.

The x and y params are provided because our search form uses an `<input type='image'>`, we could switch to an `button` input, but recommend making those design kind of changes in a separate PR

assets/js/app.js
AMZN-alexpete: We could if the input field is empty. I'd be fine with it either way.  It doesn't appear that the visible input field is hurting the UX currently - I checked on mobile and it doesn't appear to cover anything up when it's visible.

layouts/partials/search-results.html
```diff
@@ -1,5 +1,23 @@
...
+    $('#search-title').text('Searching for ' + '"' + queryParam + '" ...');
+    $("#search-input").val(queryParam);
+  }
+  // close off this script block so it gets evaluated before the heavy lifting
+</script>
 <script>
-  window.store = {
+  if(queryParam !== null && queryParam.length > 0)
+  {
+    window.store = {
```
AMZN-alexpete, line 20: We should make a future task to avoid building the index every time and do something more like this:  https://palant.info/2020/06/04/the-easier-way-to-use-lunr-search-with-hugo/

layouts/partials/search-results.html
```diff
@@ -11,15 +29,51 @@
           "url": "{{ .Permalink }}"
       },
       {{ end }}
-  }
+    }
 
+    const idx = lunr(function () {
```
AMZN-alexpete, line 34: FWIW it appears this code was heavily inspired by https://victoria.dev/blog/add-search-to-hugo-static-sites-with-lunr/#search-for-the-query-with-lunr

sptramer (approved):

PR #1338, opened 2022-01-04
Remove unused hidden duplicated nav menu content

AMZN-alexpete:
## Change summary
Every docs page contained the full nav menu for every docs subsection, and the docs landing page also included the entire nav menu in hidden HTML for mobile.  This change removes that unused duplicated HTML and should help with our SEO issues, making each page look less like the landing page to crawlers and help with bandwidth by removing a large chunk of unused HTML.

- no changes in UX or functionality
- remove all but the top level nav sub sections from the docs landing page for mobile.
- remove the hidden navigation menu subsections.  For example, in the /docs/welcome-guide, don't include the nav menu items for /docs/release-notes or /docs/user-guide

### DIFF
--- a/layouts/docs/docs_landing.html
+++ b/layouts/docs/docs_landing.html
@@ -3,6 +3,11 @@
 {{ $allSections  := site.Sections }} <!-- "blog" and "docs" -->
 {{ $docsSections := where $allSections "Section" "docs" }}
 
+{{/* Include the top level navigation for mobile */}}
+<section class="hide-content docs-sidebar">
+  {{ partial "section-nav.html" . }}
+</section>
+
 <div class="container pt-3 px-5 pb-3">
   <div class="row flex">
     <div class="col-12 docs-content">
--- a/layouts/docs/docs_landing.html
+++ b/layouts/docs/docs_landing.html
@@ -3,10 +3,6 @@
 {{ $allSections  := site.Sections }} <!-- "blog" and "docs" -->
 {{ $docsSections := where $allSections "Section" "docs" }}
 
-<section class="hide-content docs-sidebar">
-  {{ partial "section-nav.html" . }}
-</section>
-
 <div class="container pt-3 px-5 pb-3">
   <div class="row flex">
     <div class="col-12 docs-content">
--- a/layouts/partials/section-nav.html
+++ b/layouts/partials/section-nav.html
@@ -10,8 +10,13 @@
 <nav class="docs-nav" id="docs-nav">
   <ul>
     {{ range .FirstSection.Pages }}
-      {{/* Only display pages inside the current /section/subsection/ e.g. /docs/welcome-guide/ or /docs/release-notes/ */}}
-      {{- if $.IsDescendant . -}}
+      {{- if eq $ .FirstSection -}}
+      {{/* Display the main subsections on the docs page (for mobile) */}}
+      <li>
+          <a href="{{ .RelPermalink }}" class="docs-link">{{ .LinkTitle }}</a>
+      </li>
+      {{- else if $.IsDescendant . -}}
+      {{/* Only display pages inside the current subsection */}}
       <li>
           <a href="{{ .RelPermalink }}"
           class="docs-link{{ if .Pages }} has-children{{ end }}">
--- a/layouts/partials/section-nav.html
+++ b/layouts/partials/section-nav.html
@@ -1,6 +1,3 @@
-{{ $allSections  := site.Sections }} <!-- "blog" and "docs" -->
-{{ $docsSections := where $allSections "Section" "docs" }}
-
 <form id="search"
     action='{{ with .GetPage "/search" }}{{.Permalink}}{{end}}' method="get" class="align-items-center justify-content-center">
     <div class="search-input-view">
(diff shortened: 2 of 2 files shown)


layouts/docs/docs_landing.html
```diff
@@ -3,10 +3,6 @@
 {{ $allSections  := site.Sections }} <!-- "blog" and "docs" -->
 {{ $docsSections := where $allSections "Section" "docs" }}
 
-<section class="hide-content docs-sidebar">
-  {{ partial "section-nav.html" . }}
-</section>
-
```
AMZN-alexpete, line 9: on further testing I'm seeing on mobile we use the header menu to show the entire expandable nav and this change breaks that functionality.

AMZN-daimini:
Note that this changes the contents of the mobile menu dramatically.

This is the mobile menu on the docs landing page on current o3de.org
image
Note that each section has an arrow, and can be expanded
image

After this change, the mobile menu on the landing page will only show the top categories, with no way to expand them
image

In internal docs pages, the mobile menu currently shows all categories, including other top categories. This is because the mobile menu is the only navigation, the menu on the top right is completely hidden on mobile
image

This change makes it so that the current top category is the only one shown, meaning that there's no way to switch to a different top category on mobile (if not through clicking on the logo in the header to go back to home, then click learn, then use the menu in that landing page to select the category).
image


I think these changes actively make navigation harder for users, and honestly I'm not even sure they will solve the SEO issues to begin with - as far as I know, google crawlers can identify elements that are hidden via display: none and take that aspect into account.
Even if my assumption is wrong, though, I believe we should look for different ways to address the issue.

AMZN-alexpete:
@AMZN-daimini ,  you're right that this change would definitely remove the ability to navigate from any doc page to any other doc page in the docs site on mobile using that full navigation menu.  The reason I thought this might be an acceptable change is because it makes the mobile site navigation more like the desktop site navigation limitations - you don't see the full nav menu on every page, just links to the top sections, and then once you're in a section we hide the nav for every other docs section.  Personally I prefer a full navigation menu, but would be willing to compromise especially if it fixes our Google search index issues.  One key piece of data I'm missing here is analytics - do we even have any users browsing the docs on mobile? Is it a small or large percent?  Are they navigating with the nav menu?

Regarding the SEO change, I recommend we hold off on reviewing/submitting this change for a week so we can give the Google webcrawlers a chance to update the index on the docs site.  Maybe the title and search changes we made will be enough to resolve the issue, but if not they maybe we can revisit this change as something we could try temporarily to make our doc pages not look so similar.  I have noticed that `canonical` page `/docs`  which Google thinks is the same as all our other docs pages was crawled by Google in mobile mode `Crawled as  Googlebot smartphone`.

AMZN-daimini:
> One key piece of data I'm missing here is analytics - do we even have any users browsing the docs on mobile? Is it a small or large percent? Are they navigating with the nav menu?

Yeah we don't really have that kind of data right now AFAIK. My suggestion is mostly to discuss this change in terms of website UX before implementing it.

> Regarding the SEO change, I recommend we hold off on reviewing/submitting this change for a week so we can give the Google webcrawlers a chance to update the index on the docs site. Maybe the title and search changes we made will be enough to resolve the issue, but if not they maybe we can revisit this change as something we could try temporarily to make our doc pages not look so similar. I have noticed that `canonical` page `/docs` which Google thinks is the same as all our other docs pages was crawled by Google in mobile mode `Crawled as Googlebot smartphone`.

Makes sense. I think it should also be possible to mark the menus so that the crawler does not see them, and that may serve the same purpose without making navigation more complicated for mobile users.

sptramer:
@o3de/sig-ui-ux Tagging in UI/UX since that's the most important change here. We still don't have full user agent/device analytics, so it's hard to say if limiting mobile dev temporarily to handle this issue is a good tradeoff.

sptramer:
@amzn-leenguy @yuyihsu - Need a UI/UX eval on this since it changes navigation. I'm OK with it.

sptramer (approved):
LGTM and makes the website load dramatically faster (even over a fiberop cable).

amzn-leenguy:
Hello @AMZN-alexpete , @AMZN-daimini and @sptramer . Apologize for being late to the party. 

I am on the same page here as @AMZN-daimini in that we shouldn't remove the ability for people to navigate for the sake of SEO. Here is my middle ground suggestion:

- If we move forward with Alex's suggestion: Are we able to add an option on the mobile experience to have: 

< Back

(in the mobile nav menu above all the current sub pages of the page your in)

which would load the following options

Screen Shot 2022-02-22 at 11 33 19 AM

This way we aren't completely just blocking the user to go up a tier. 

I checked out the netlify link here: https://deploy-preview-1338--o3deorg.netlify.app/docs/learning-guide/
And I can't agree with this. (especially because we haven't pushed the change to make clicking on the logo go back to docs home yet. But I know @chanmosq fixed this recently)

chanmosq:
> I checked out the netlify link here: https://deploy-preview-1338--o3deorg.netlify.app/docs/learning-guide/
And I can't agree with this. (especially because we haven't pushed the change to make clicking on the logo go back to docs home yet. But I know @chanmosq fixed this recently)

@amzn-leenguy @AMZN-daimini @AMZN-alexpete --
The O3DE Learn logo now links back to the documentation landing page [PR #1412](https://github.com/o3de/o3de.org/pull/1412). This is one way to go back to the other guides on mobile view. Alex may need to pull the latest changes to reflect that behavior in the netlify website preview.

AMZN-alexpete:
@amzn-leenguy because our site doesn't do any special redirects for mobile, you can test most UX by resizing your browser window down to the size of a mobile device.  @chanmosq no doubt you are correct, this PR hasn't merged latest in a while.

chanmosq:
@amzn-leenguy @AMZN-daimini Can we get some feedback or approval on this?

AMZN-daimini:
Hey, @amzn-leenguy and I discussed this again today, but honestly I'm still of the opinion that this change hinders user navigation from mobile, and is not the most effective way of solving the underlying problem. If the navigation menu being in all pages indeed created SEO issues, there are more effective ways of hiding it from robots/crawlers than removing menu items and forcing users to load multiple places to reach their intended destination.

chanmosq:
@AMZN-alexpete Given the feedback, can we close this and consider another solution?

AMZN-alexpete:
> @AMZN-alexpete Given the feedback, can we close this and consider another solution?

What other solution did you have in mind?

AMZN-daimini:
First of all, it this SEO concern still an issue?

> Regarding the SEO change, I recommend we hold off on reviewing/submitting this change for a week so we can give the Google webcrawlers a chance to update the index on the docs site. Maybe the title and search changes we made will be enough to resolve the issue, but if not they maybe we can revisit this change as something we could try temporarily to make our doc pages not look so similar. I have noticed that canonical page /docs which Google thinks is the same as all our other docs pages was crawled by Google in mobile mode Crawled as Googlebot smartphone.

If so, what are the exact issues? Can we get some analytics from google? The navigation is marked as `<nav>` so I'd like to know more about why we think navigation makes pages look similar to the crawler. Maybe we can improve the situation by improving the tagging for the actual content?

sptramer:
Closing PR as we need to discuss alternative solutions. This would likely need to involve a working group under sig-docs-community to evaluate/create work issues for any implementation for better nav support. @chanmosq has also partially addressed this issue.

PR #1361, opened 2022-01-20
Added docs for Bloom Component

chanmosq:
### Submission Checklist:

* [X] **Consistency** - Does the content consistently follow the [style guide](https://o3de.org/docs/contributing/to-docs/style-guide/)?


content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
 ## Properties
 
-{{< todo issue=https://github.com/o3de/o3de.org/issues/717 >}}
-Add properties.
-{{< /todo >}}
\ No newline at end of file
+![Bloom component interface](/images/user-guide/components/reference/atom/bloom-component-ui.png)
+
+
+### Base properties
```
FiniteStateGit, line 26:
```suggestion
### Bloom properties
```
This could just be 'bloom' since there are no other properties sections.

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+
+| Property | Description | Value | Default |
+| - | - | - | - |
+| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | The default value |
```
FiniteStateGit, line 30:
```suggestion
| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | `Enabled` |
```

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+| **Threshold** | Bloom effect applies only to pixels with a brightness value greater than this threshold. | Float: 0.0 to Infinity  | `1.0` |
+| **Knee** | Creates a gradual transition between the pixels below and above the threshold. This softens and spreads the bloom effect.  | Float: 0.0 to 1.0 | `0.5` |
+| **Intensity** | Scales the intensity of the bloom effect. | Float: 0.0 to Infinity | `0.5` |
+| **Enable Bicubic** | If enabled, applies bicubic filtering. This can help reduce undesired artifacts that may appear. | Boolean |  False |
```
FiniteStateGit, line 34:
```suggestion
| **Enable Bicubic** | If enabled, applies bicubic filtering. This can help reduce undesired artifacts that may appear. | Boolean |  `Disabled` |
```

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+| **Kernal Size Scale** | Scales the size of the kernal. | Float: 0.0 to 2.0 | `1.0` |
```
FiniteStateGit, line 35:
```suggestion
| **Kernel Size Scale** | Scales the size of the kernel. | Float: 0.0 - 2.0 | `1.0` |
```

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+| **Kernal Size 0 to 4** | Smooths the kernal size by a percentage of the render target's width. For example, a kernal size of `0.04` translates to a 4-by-4 kernal on a 100-by-100 pixel image and a 40-by-40 kernal on a 1000-by-1000 pixel image. Larger values causes more scattering at a higher cost of computation. The maximum kernal size is 128. | Float: 0.0 to 1.0 | Kernal Size<br><ul><li>0: `0.04`</li><li>1: `0.08`</li><li>2: `0.16`</li><li>3: `0.32`</li><li>4: `0.64`</li></ul> |
```
FiniteStateGit, line 36:
```suggestion
| **Overrides - Enabled Override** | If `Enabled`, all Bloom component properties will be set to the values specified in the **Overrides** property group. | Boolean | `Enabled` |
| **Kernel Size 0 to 4** | Smooths the kernel size by a percentage of the render target's width. For example, a kernel size of `0.04` translates to a 4-by-4 kernel on a 100-by-100 pixel image and a 40-by-40 kernel on a 1000-by-1000 pixel image. Larger values causes more scattering at a higher cost of computation. The maximum kernel size is 128. | Float: 0.0 - 1.0 | Kernel Size<br><ul><li>0: `0.04`</li><li>1: `0.08`</li><li>2: `0.16`</li><li>3: `0.32`</li><li>4: `0.64`</li></ul> |
```
Added a suggested row to document the **Overrides** property group.  It does look like every property is overridden (and maxed out) by default! Changed spelling of kernel and added hyphen to value range in the line that was moved.

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+| **Tint 0 to 4** | Adds a color tint to each bloom stage. Bloom stages are additively blended to produce the final result. | Vector3: 0 to 255 | X: `255`. Y: `255`, Z: `255` |
```
FiniteStateGit, line 37:
```suggestion
| **Tint 0 to 4** | Adds a color tint to each bloom stage. Bloom stages are additively blended to produce the final result. | Vector3: 0 - 255 | X: `255`. Y: `255`, Z: `255` |
```

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,38 @@
...
+![Bloom component interface](/images/user-guide/components/reference/atom/bloom-component-ui.png)
+
+
+### Base properties
+
+| Property | Description | Value | Default |
+| - | - | - | - |
+| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | The default value |
+| **Threshold** | Bloom effect applies only to pixels with a brightness value greater than this threshold. | Float: 0.0 to Infinity  | `1.0` |
+| **Knee** | Creates a gradual transition between the pixels below and above the threshold. This softens and spreads the bloom effect.  | Float: 0.0 to 1.0 | `0.5` |
```
FiniteStateGit, line 32:
```suggestion
| **Knee** | Creates a gradual transition between the pixels below and above the threshold. This softens and spreads the bloom effect.  | Float: 0.0 - 1.0 | `0.5` |
```
I've been using a hyphen for ranges that use two numeric values and 'to' when the range includes a word instead of a number.  I don't have a strong opinion about it, but I'll include the edit if you want.

FiniteStateGit (changes requested):

content/docs/user-guide/components/reference/atom/bloom.md
chanmosq: Adding in. You summarized Overrides well, thanks!

FiniteStateGit (approved):

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,39 @@
...
+
+### Bloom properties
+
+| Property | Description | Value | Default |
+| - | - | - | - |
+| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | `Enabled` |
+| **Overrides** | If enabled, all Bloom component properties will be set to the values specified in the **Overrides** property group. | Boolean | `Enabled` |
+| **Threshold** | Bloom effect applies only to pixels with a brightness value greater than this threshold. | Float: 0.0 to Infinity  | `1.0` |
+| **Knee** | Creates a gradual transition between the pixels below and above the threshold. This softens and spreads the bloom effect.  | Float: 0.0 - 1.0 | `0.5` |
+| **Intensity** | Scales the intensity of the bloom effect. | Float: 0.0 to Infinity | `0.5` |
```
willihay, line 34: Happened to be in the Editor, so I loaded up the component and noticed that Intensity has a slider bar.
```suggestion
| **Intensity** | Scales the intensity of the bloom effect. | Float: 0.0 to 10000.0 | `0.5` |
```

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,39 @@
...
\ No newline at end of file
+![Bloom component interface](/images/user-guide/components/reference/atom/bloom-component-ui.png)
+
+
+### Bloom properties
+
+| Property | Description | Value | Default |
+| - | - | - | - |
+| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | `Enabled` |
+| **Overrides** | If enabled, all Bloom component properties will be set to the values specified in the **Overrides** property group. | Boolean | `Enabled` |
```
willihay, line 31: I don't get it. Why are overrides on by default? Seems like a strange UI/UX choice to me.

content/docs/user-guide/components/reference/atom/bloom.md
```diff
@@ -1,20 +1,39 @@
...
-{{< /todo >}}
\ No newline at end of file
+![Bloom component interface](/images/user-guide/components/reference/atom/bloom-component-ui.png)
+
+
+### Bloom properties
+
+| Property | Description | Value | Default |
+| - | - | - | - |
+| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | `Enabled` |
```
willihay, line 30: Oddly enough, when I first add the Bloom component, it's disabled.
```suggestion
| **Enable Bloom** | If enabled, activates the bloom effect. | Boolean | `Disabled` |
```

content/docs/user-guide/components/reference/atom/bloom.md
willihay: Also, it doesn't seem to be a simple value override to me. Looking at Intensity, for example: The normal range is 0 to 10000. (Anything above 2 produces absurd-looking results, but whatever.) If I set Intensity to 2.5 and the Intensity Override factor to 0.0, you'd think that I'd see a difference when I toggle Enable Override, but I don't. Also, the scale of the Intensity Override is only 0 to 1. If it were a simple value replacement override, the scale should be the same.

willihay (changes requested):

content/docs/user-guide/components/reference/atom/bloom.md
chanmosq: I thought this was odd too. Thanks for catching this!

content/docs/user-guide/components/reference/atom/bloom.md
FiniteStateGit: Some UI sliders have soft min and max values.  A user can type a value that is smaller or larger than dragging the slider would allow.  The slider's range is supposed to be set to sensible values, which may not be the case here.

content/docs/user-guide/components/reference/atom/bloom.md
FiniteStateGit: https://github.com/o3de/o3de/issues/5993

FiniteStateGit: An open issue related to the intensity slider: https://github.com/o3de/o3de/issues/5993.

PR #1464, opened 2022-03-07
Update API reference to correct product version and fix broken Gem generation

sptramer:


sptramer:
Reviewers - you obviously don't need to check every individual file. Just do some quick spot-check to make sure that the update is consistent and correct before merge.

willihay:
Wow, 18 hours later and the deploy preview is still processing? Hmm... is that normal for a change of this size? Is processing time size-dependent somehow?

sptramer:
It may have stalled out in actions somewhere - the deploy shows as completed (https://app.netlify.com/sites/o3deorg/deploys/6226970df1245b00088e49d2) but isn't deployed. This is a pretty frequent problem with API updates because they're such enormous changes, even if only dates are updated (which we should discuss potentially pruning out).

sptramer:
Forcibly restarted the deploy, let's see if it works this time. If it doesn't, then we have a pretty serious infrastructure problem to work out!

FiniteStateGit (approved):

micronAMZN (approved):

PR #1495, opened 2022-03-18
Website navigation experience: Separate section’s group title and page title

chanmosq:
### Problem

A *section* in the docs contains at least an `_index.md` and other single pages or nested sections.  In the section’s `_index.md` file, the `linkTitle` parameter defines the section’s navigation group title and page title on the website’s *doc section navigation* on the left-hand side (see image below). The problem with this is you can’t rename the the page title to a fitting name such as “Overview”. 

image

### Past solution

*I investigated this problem earlier in the year and tried the following solution, but it’s not the cleanest.* 

For a sections in the docs, I removed the contents of `_index.md` file except for the metadata, so it doesn’t render on the doc section navigation in the website. Instead, I put those contents in another page (e.g. `[overview.md](http://overview.md)`) and set that page as the first page of the section. The problem here is that the section’s page, a commonly visited page by users, will be empty. You can see this in the example below, where the first page of the Gem’s section is successfully titled “Overview”, but directs you to the `docs/user-guide/gems/overview/` page (see A), leaving the `docs/user-guide/gems/` page empty (see B). 

Even if all links on the website direct users to the separate overview page `docs/user-guide/gems/overview/` instead of the section’s landing page. `docs/user-guide/gems/`, the section’s landing page still gets rendered. So it’s very likely for users to chance upon that page and see that it’s empty, causing confusion in their experience with the website. An empty page is especially undesirable when the website is viewed in “mobile” mode, because the doc section navigation is hidden by default, making it even less likely for users to find section topics. 

**A**

image

**B**

image

### Proposed solution

This is my new proposed solution. 

I introduce a new variable in Hugo pages, `.sectionPageTitle`, where you can define the section’s page title by including it in the metadata. Then, in the `section-nav-walk.html` file, I added a functionality so that if the page is a section, it’s page title will appear on the doc section navigation with the name provided by the `.sectionPageTitle` parameter, if one is provided. The section’s group title in the doc section navigation remains set by the `linkTitle` parameter, as usual.

In the example below, the `docs/user-guide/editor/_index.md` page contains the following metadata and renders the website as shown in the image after: 

```markdown
---
linkTitle: Editor
title: O3DE Editor
sectionPageTitle: Overview
weight: 300
---
```
image

### Other notes

- Because of the logic in `section-nav-walk.html` , any page that isn’t a section (doesn’t contain `_index.md`) is not affected by a `.sectionPageTitle` parameter in the metadata.
- The `.sectionPageTitle` parameter also applies to high-level sections, such as **User Guide**, **Contributing**, **Tutorials**, and so on.
- `.sectionPageTitle` is a user-created page variable (that I created for the purpose here) and can be accessed using the [`.Params` function](https://gohugo.io/functions/param/#readout).

## Submission Checklist:

### DIFF
--- a/layouts/partials/section-nav-walk.html
+++ b/layouts/partials/section-nav-walk.html
@@ -4,15 +4,17 @@
         {{ $isHere := in $here .RelPermalink }}
         <li>
             <a href="{{ .RelPermalink }}"
-            class="docs-link{{ if .Pages }} has-children{{ end }}">
-            {{ .LinkTitle }}</a>
+            class="docs-link{{ if .Pages }} has-children{{ end }}"> 
+                {{.LinkTitle}} 
+            </a>
             {{ if .Pages }}
                 <ul>
                     {{ if and ( .IsSection ) (ge (len .Content) 1) }}
                     <li>
                         <a href="{{ .RelPermalink }}"
-                        class="docs-link">
-                        {{ .LinkTitle }}</a>
+                        class="docs-link">                        
+                            {{ if .Params.sectionPageTitle }} {{.Params.sectionPageTitle}} {{else}} {{.LinkTitle}} {{end}}
+                        </a>
                     </li>
                     {{ end }}
                     {{ partial "section-nav-walk.html" . }}


willihay (approved):
Approving because this is definitely an improvement over what we have now, even though it's still not the ideal that I'd like to see. If I understand correctly, essentially what we're gaining here is the ability to rename the _index page so that you have the option of giving it a different name than the section, so that you're not just repeating the section name when you open the collapsible section.

Ideally, I'd prefer to see the ToC navigation work the way it did for the Lumberyard docs, where you could click on the section name and it would open to the _index page _and_ expand the section. Did you also look into a solution like that?

sptramer:
@o3de/sig-ui-ux - You don't have a reviewer group but we would like UI/UX to provide feedback on this issue before merge!

amzn-leenguy:
300% approve this. Well done Chanelle on the new proposed solution.

willihay:
@sptramer UI-UX approves. Need signoff from second reviewer to merge.

sptramer:
This is a good stopgap solution - will there be a followup PR or issues file to rename/move sections around appropriately? We should no longer need `overview.md` files distinct from `_index.md` if we can do this kind of thing now.

sptramer (approved):

chanmosq:
> Ideally, I'd prefer to see the ToC navigation work the way it did for the Lumberyard docs, where you could click on the section name and it would open to the _index page _and_ expand the section. Did you also look into a solution like that?

@willihay This is another great solution. I haven't looked into it yet. The reason I went this route is because of ease and my knowledge in changing/using variables with Hugo. I'm very open to revisiting better solutions in the future.

> This is a good stopgap solution - will there be a followup PR or issues file to rename/move sections around appropriately? We should no longer need overview.md files distinct from _index.md if we can do this kind of thing now.

@sptramer I'll create an issue for this!

chanmosq:
@sptramer Created issue here: https://github.com/o3de/o3de.org/issues/1523

PR #1498, opened 2022-03-18
Updates to Debugging Asset Problems

AMZN-stankowi:
Current content is ready for review. Some sections are more stubbed out than fully featured, like performance. I figure we can expand this over time, I don't want to spend too long on one doc update right now.

Signed-off-by: AMZN-stankowi <4838196+AMZN-stankowi@users.noreply.github.com>

### Submission Checklist:

### DIFF
--- a/content/docs/user-guide/assets/asset-processor/debugging.md
+++ b/content/docs/user-guide/assets/asset-processor/debugging.md
@@ -8,16 +8,7 @@ toc: true
 
 If you are having issues with **Asset Processor**, use the methods below to debug the issues.
 
-## Who normally needs to debug Asset Processor issues?
-
-* Content creators looking into problems with their content not behaving in-engine like they expect.
-* Technical artists, assistant content creators or working on expanding their team's pipeline with Python based automation.
-* Engineers assisting content creators or technical artists with debugging those issues.
-* Engineers working on an asset builder, either changing an existing one or making a new one.
-* Engineers working on packaging content for a bundled release build of a project, using the Asset Processor to discover a gap or other issue that was found during the bundling process.
-* Engineers working on the Asset Processor itself, including the logic for the Asset Processor to communicate with the Editor or launchers.
-
-## What kinds of problems need to be solved within the Asset Processor?
+## What kinds of problems can be encountered within Asset Processing?
 
 * After changing a source asset, it takes too long for assets to finish processing.
 * The product assets don't behave like expected. They may be missing entirely, missing information, have incorrect information.
@@ -28,7 +19,7 @@ If you are having issues with **Asset Processor**, use the methods below to debu
 
 ## What are common Asset Processor related problems, and paths to address those problems?
 
-### Product Asset Thrash - Non Deterministic Product Assets
+### Non Deterministic Product Assets - Different Product Assets with No Source Changes
 
 #### Situation
 With no changes to the code for a builder, and no changes to a source asset, when that job is run multiple times, the contents of product assets end up different than the previous run.
@@ -43,19 +34,23 @@ Extraneous information is included in the product asset, such as a timestamp. Lo
 In most cases this will require an engineer to make changes to the internal builder logic to stabilize the product output.
 
 #### Debugging
-Your team most likely encounter this issue when generating [Asset Bundles](/packaging/asset-bundler/), which is the primary system it impacts. The Asset Bundling process is generally done by an engineer, who will likely work backward from first finding a product asset in bundles that was not expected to be there, because no changes were done that should have resulted in a new product asset.
+Your team most likely encounter this issue when generating [Asset Bundles](/packaging/asset-bundler/), which is the primary system non-deterministic Product Assets impacts. If a Product Asset shows up in a bundle that you didn't expect to be there because the Source Asset and associated Asset Builder did not change, then debugging tends to start by working backward from there.
 
-This thrash will be introduced by the asset builder that created this product asset, so to identify the cause you'll want to look for this product asset in the Asset Processor's asset tab, or directly in the Asset Database using an external database browsing tool. From there you can find the job that created that product asset. Once you identify the job that created this product asset, you can then find the builder that produces that job, and go to the Process Jobs function for that builder, to find where it outputs the product asset. Working backward from here, you should be able to find the cause of the thrash by exploring the code. You can also try [debugging the Asset Builder](#DebugAssetBuilders) processing that job, to step through.
+Non-deterministic behavior in Product Assets will be introduced by the asset builder that created this product asset, so to identify the cause you'll want to look for this product asset in the Asset Processor's asset tab, or directly in the Asset Database using an external database browsing tool. From there you can find the job that created that product asset. Once you identify the job that created this product asset, you can then find the builder that produces that job, and go to the Process Jobs function for that builder, to find where it outputs the product asset. Working backward from here, you should be able to find the cause of the Product Asset changing by exploring the code. You can also try [debugging the Asset Builder](#DebugAssetBuilders) processing that job, to step through.
 
-To help identify the block of logic in the product asset that is thrashing, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
+To help identify the block of logic in the Product Asset that is changing each time it is processed, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
 
-### Product Asset Sub ID Thrash - Same product generates different Sub ID
+Another debugging option is setting up automation to catch this early. If you have an automated system that processes assets for your project, you can expand this automation to also 
+
+### Same Product Asset generated with different Sub ID
 
 #### Situation
-With no changes to the code for a builder, and no changes to a source asset, when that job is run multiple times, the Sub IDs of product assets end up different than the previous run.
+The Sub ID for a Product Asset changes when it is not expected to change.
+
+In the worst case, the Sub ID is changing when no changes have been made to the Source Asset or builder that created the Product Asset. In other cases, a change to the Source Asset results in a Product Asset that has a different Sub ID than it did previously.
 
 #### Effect
-Assets are mostly referenced by asset ID, the UUID of the source asset with the Sub ID for that product asset. If this is not consistent and the Sub ID thrashes, then external references to this asset can end up incorrect, pointing at a different asset after a rebuild or pointing at no asset if no other products use that Sub ID.
+Assets are mostly referenced by asset ID, the UUID of the source asset with the Sub ID for that product asset. If this is not consistent and the Sub ID changes for a Product Asset when it should not change, then external references to this asset can end up incorrect, pointing at a different asset after a rebuild or pointing at no asset if no other products use that Sub ID.
 
 #### Example
 Here is an example with scene files and prefabs:
@@ -72,12 +67,12 @@ Here is an example with scene files and prefabs:
 
 This happens because scene processing tracks the meshes by name, and by swapping the names between the meshes you've replaced which product assets are which. The prefab tracks the outgoing product asset reference via asset ID. When I tested this locally, my Mesh_A azmodel had the Product Asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:270123712, and Mesh_B has the asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:285065766. When I swapped the FBX files, the node within the file named Mesh_A still had that same first ID, but the geometry was a cylinder now, changing what it looked like in the prefab.
 
-Note that in this case, this behavior is 100% intentional. There is no bug in this example, but other times this thrash occurs it may be considered a bug.
+Note that in this case, this behavior is 100% intentional. There is no code issue in this example. In this case, if this change was not intended by the FBX author, it is a content bug. The following sections will cover how to identify and correct issues like this.
 
 #### Common Causes
 
 ##### Change to source asset causes Sub IDs to shift
-The most common place Sub ID thrash occurs is when the Sub ID generation is based on data within the source asset, and that data can change. For example, if a node name is hashed to generate the Sub ID for a product asset, then a content creator renaming that node in the source asset will cause the associated product asset to generate a new Sub ID.
+The most common place Sub ID changes occurs is when the Sub ID generation is based on data within the source asset, and that data can change. For example, if a node name is hashed to generate the Sub ID for a product asset, then a content creator renaming that node in the source asset will cause the associated product asset to generate a new Sub ID.
 
 ##### Non-deterministic Sub ID generation from builder
 Sometimes the logic for how a builder may not be fully deterministic. For example, if a builder just increments an integer to use as the Sub ID for each product it generates, then a source asset change that results in a change in the products to remove one, will result in the Sub IDs changing for all product assets after the removed product asset.
@@ -90,8 +85,26 @@ There are times when a builder author needs to make a change to how the builder 
 ##### Stabilize Sub IDs while authoring a builder
 It is the responsibility of the person authoring a builder to define the rules for how Sub IDs are generated by that builder. When doing so, we recommend putting in effort to stabilize Sub ID generation as much as possible: Think about how the source asset can be modified, and how you can maintain continuity in product asset Sub ID generation based on those changes.
 
+Tips for stabilizing Sub IDs for Product Assets authoring a builder:
+* A randomly generated Sub ID is usually going to result in the Sub ID changing when not expected, even if the generation is semi-randomized. Randomly generated Sub IDs should be avoided.
+* Often times, the obvious first way to assign Sub ID for Product Assets is based on the ordered index of that information in the Source Asset. Often times this results in problems with Sub IDs changing, because changes in the Source Asset can result in the ordered list used to shuffle. For example if the content creator removes an entry and causes everything else to shuffle down an index.
+* Hashing the name of an object to use as a reference can work as long as content creators don't expect to rename these objects have have the Sub ID remain stable.
+   * Hashing the name of an object to use as the Sub ID can also be beneficial to content creators in that it allows them to replace content by creating new data with the same name to generate a Product Asset with the same ID.
+* Sometimes there may be a stable internal ID you can use, that content creators do not get direct control over. For example, there may be an internal node structure that generates unique internal IDs for nodes that content creators cannot change. This may be more stable in some ways to use to generate Sub IDs, but consider your content creator's workflows before committing to using this for Sub ID generation. There may be times a content creator specifically wants to make a new thing to replace something, and have the new Product Asset seamlessly be used everywhere the old one was. See the previous point about using object names for an example.
+
+
 ##### Learning how Sub ID generation works for builders that process content you author
-This is preventing this problem more than it is a solution. Much of the content generated for O3DE projects is going to be referenced and used by other content. When you initially sit down to author your content, think about what the output should look like, and how it will be used. When you make changes to the source asset, also think about how that will effect the product assets generated, and how you can maintain continuity with other assets that reference that content. For example, if you're responsible for an FBX file that outputs multiple azmodel product assets, and you want to make a change to this FBX to remove or rename one of those product assets, first you need to look up all references to that model, so you can figure out how to update those. If several prefabs reference the model you are going to remove, then those prefabs may not function as expected, so you'll want to update those first to be ready for your change to the FBX file, so it won't break those prefabs.
+As a content creator, learning how Sub IDs are generated for Product Assets created from the Source Assets you work on can prevent these problems in the first place.
+
+When a Sub ID changes for a Product Asset, it will cause all references to that Product Asset to break, because the Asset ID will no longer resolve to that asset.
+
+During initial content creation, especially for scene files like FBX, thinking through how you want that content referenced elsewhere can help keep the Sub ID for the Product Assets stable.
+
+If you need to make a change to the Source Asset, keep in mind how your changes will impact the Sub IDs of the Product Assets. If you do need to change the Product Assets, a good pattern to follow is
+1. Create a stub of the new Product Asset, or leave a stub for the old Product Asset.
+1. Find all references to the old Product Asset.
+1. Update those as necessary, working with other members on your team to do so.
+1. Once you're sure nothing is left referencing the old Product Asset, then make the final change to remove it from the Source Asset.
 
 ##### Fix the broken references - update the source asset
 Updating the source asset in a way that it will output a product asset with the same sub ID will fix any broken connections. This works best when the product asset was removed unintentionally, either the content creator was unaware the product asset was in use, or they made a change to the source asset not realizing it would cause this issue.
@@ -99,12 +112,19 @@ Updating the source asset in a way that it will output a product asset with the 
 ##### Fix the broken references - update the content referencing the missing product asset
 If the removal of the product asset was intentional, or the source asset cannot safely be updated to restore the product asset, it may be necessary to instead update everything referencing the broken product asset. In this case you'll need to search for all broken references, and for each type or content or code referencing the missing product asset, update it within the relevant system to handle the removal of this product asset or change to the Sub ID.
 
+For example, if an FBX file was updated to no longer generate a Product Asset, and an O3DE Prefab referenced this product asset via a mesh component on an entity, you can use the prefab editing workflow to fix this.
+1. Load the prefab to be edited.
+1. Find the component with the broken reference.
+1. Either change it to reference alternate content, or remove the component/entity.
+
 #### Debugging
 If content appears missing or broken, try opening the file that references broken content in a text editor or another editor that might show additional details. Search for the broken reference, it will either be an asset ID which is a combination UUID and Sub ID, or it will be a reference via relative path.
 
 If you can find the missing reference and it is an Asset ID reference, then you can use the UUID to determine the Source Asset. From here, you can look at the history of this Source Asset in your source control to see what changes may have come in, as well as who authored those changes. You can experiment by rolling back the file locally to see if the missing product asset shows up again.
 
-If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where this thrash is coming from.
+If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where the change in Sub IDs is coming from.
+
+If the builder is an O3DE builder, and not one that your team has created, you can create a ticket [here](https://github.com/o3de/o3de/issues) to get this looked at.
 
 ## View Asset Processor Logs 
 
--- a/content/docs/user-guide/assets/asset-processor/debugging.md
+++ b/content/docs/user-guide/assets/asset-processor/debugging.md
@@ -8,6 +8,103 @@ toc: true
 
 If you are having issues with **Asset Processor**, use the methods below to debug the issues.
 
+## Who normally needs to debug Asset Processor issues?
+
+* Content creators looking into problems with their content not behaving in-engine like they expect.
+* Technical artists, assistant content creators or working on expanding their team's pipeline with Python based automation.
+* Engineers assisting content creators or technical artists with debugging those issues.
+* Engineers working on an asset builder, either changing an existing one or making a new one.
+* Engineers working on packaging content for a bundled release build of a project, using the Asset Processor to discover a gap or other issue that was found during the bundling process.
+* Engineers working on the Asset Processor itself, including the logic for the Asset Processor to communicate with the Editor or launchers.
+
+## What kinds of problems need to be solved within the Asset Processor?
+
+* After changing a source asset, it takes too long for assets to finish processing.
+* The product assets don't behave like expected. They may be missing entirely, missing information, have incorrect information.
+* Finding and attempting to fix a warning, error, or failure that occured during asset processing for a particular job.
+* Asset packaging isn't working as expected, content is missing or content is being included in bundles that was not expected to be there.
+* Assistant while working on a builder to add a new feature, fix a bug, or create a new builder.
+* Different members of a team getting different results from asset processing, impacting the ability for the team to cooperate.
+
+## What are common Asset Processor related problems, and paths to address those problems?
+
+### Product Asset Thrash - Non Deterministic Product Assets
+
+#### Situation
+With no changes to the code for a builder, and no changes to a source asset, when that job is run multiple times, the contents of product assets end up different than the previous run.
+
+#### Effect
+More assets than expected will show up when attempting to generate Delta Asset Bundles to create content patches for your game, causing your players to download more content than necessary.
+
+#### Common Causes
+Extraneous information is included in the product asset, such as a timestamp. Logic in the builder is non-deterministic.
+
+#### Solutions
+In most cases this will require an engineer to make changes to the internal builder logic to stabilize the product output.
+
+#### Debugging
+Your team most likely encounter this issue when generating [Asset Bundles](/packaging/asset-bundler/), which is the primary system it impacts. The Asset Bundling process is generally done by an engineer, who will likely work backward from first finding a product asset in bundles that was not expected to be there, because no changes were done that should have resulted in a new product asset.
+
+This thrash will be introduced by the asset builder that created this product asset, so to identify the cause you'll want to look for this product asset in the Asset Processor's asset tab, or directly in the Asset Database using an external database browsing tool. From there you can find the job that created that product asset. Once you identify the job that created this product asset, you can then find the builder that produces that job, and go to the Process Jobs function for that builder, to find where it outputs the product asset. Working backward from here, you should be able to find the cause of the thrash by exploring the code. You can also try [debugging the Asset Builder](#DebugAssetBuilders) processing that job, to step through.
+
+To help identify the block of logic in the product asset that is thrashing, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
+
+### Product Asset Sub ID Thrash - Same product generates different Sub ID
+
+#### Situation
+With no changes to the code for a builder, and no changes to a source asset, when that job is run multiple times, the Sub IDs of product assets end up different than the previous run.
+
+#### Effect
+Assets are mostly referenced by asset ID, the UUID of the source asset with the Sub ID for that product asset. If this is not consistent and the Sub ID thrashes, then external references to this asset can end up incorrect, pointing at a different asset after a rebuild or pointing at no asset if no other products use that Sub ID.
+
+#### Example
+Here is an example with scene files and prefabs:
+1. Create a new FBX file named shapes_1.fbx with two meshes in it, a cube and a cylinder. Name the cube Mesh_A and the cylinder Mesh_B.
+1. Swap the names of these two meshes, so the cube is now named Mesh_B and the cylinder is named Mesh_A. Save this to a second FBX named shapes_2.fbx.
+1. Place shapes_1.fbx in your project, and rename it to shapes.fbx.
+1. Create a new prefab with an entity in it with a mesh component.
+1. Assign the cube Mesh_A from shapes.fbx to this component.
+1. Save and quit the Editor.
+1. Replace the shapes.fbx file in your project with shapes_2.fbx.
+1. Load the Editor.
+1. Load the prefab you saved earlier.
+1. Notice that the mesh on the entiyt component is still Mesh_A, but now visually looks like the cylinder.
+
+This happens because scene processing tracks the meshes by name, and by swapping the names between the meshes you've replaced which product assets are which. The prefab tracks the outgoing product asset reference via asset ID. When I tested this locally, my Mesh_A azmodel had the Product Asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:270123712, and Mesh_B has the asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:285065766. When I swapped the FBX files, the node within the file named Mesh_A still had that same first ID, but the geometry was a cylinder now, changing what it looked like in the prefab.
+
+Note that in this case, this behavior is 100% intentional. There is no bug in this example, but other times this thrash occurs it may be considered a bug.
+
+#### Common Causes
+
+##### Change to source asset causes Sub IDs to shift
+The most common place Sub ID thrash occurs is when the Sub ID generation is based on data within the source asset, and that data can change. For example, if a node name is hashed to generate the Sub ID for a product asset, then a content creator renaming that node in the source asset will cause the associated product asset to generate a new Sub ID.
+
+##### Non-deterministic Sub ID generation from builder
+Sometimes the logic for how a builder may not be fully deterministic. For example, if a builder just increments an integer to use as the Sub ID for each product it generates, then a source asset change that results in a change in the products to remove one, will result in the Sub IDs changing for all product assets after the removed product asset.
+
+##### Change in builder logic changes Sub ID generation
+There are times when a builder author needs to make a change to how the builder generates Sub IDs for product assets, resulting in all product assets output from that builder changing Sub IDs the next time they are processed.
+
+#### Solutions
+
+##### Stabilize Sub IDs while authoring a builder
+It is the responsibility of the person authoring a builder to define the rules for how Sub IDs are generated by that builder. When doing so, we recommend putting in effort to stabilize Sub ID generation as much as possible: Think about how the source asset can be modified, and how you can maintain continuity in product asset Sub ID generation based on those changes.
+
+##### Learning how Sub ID generation works for builders that process content you author
+This is preventing this problem more than it is a solution. Much of the content generated for O3DE projects is going to be referenced and used by other content. When you initially sit down to author your content, think about what the output should look like, and how it will be used. When you make changes to the source asset, also think about how that will effect the product assets generated, and how you can maintain continuity with other assets that reference that content. For example, if you're responsible for an FBX file that outputs multiple azmodel product assets, and you want to make a change to this FBX to remove or rename one of those product assets, first you need to look up all references to that model, so you can figure out how to update those. If several prefabs reference the model you are going to remove, then those prefabs may not function as expected, so you'll want to update those first to be ready for your change to the FBX file, so it won't break those prefabs.
+
+##### Fix the broken references - update the source asset
+Updating the source asset in a way that it will output a product asset with the same sub ID will fix any broken connections. This works best when the product asset was removed unintentionally, either the content creator was unaware the product asset was in use, or they made a change to the source asset not realizing it would cause this issue.
+
+##### Fix the broken references - update the content referencing the missing product asset
+If the removal of the product asset was intentional, or the source asset cannot safely be updated to restore the product asset, it may be necessary to instead update everything referencing the broken product asset. In this case you'll need to search for all broken references, and for each type or content or code referencing the missing product asset, update it within the relevant system to handle the removal of this product asset or change to the Sub ID.
+
+#### Debugging
+If content appears missing or broken, try opening the file that references broken content in a text editor or another editor that might show additional details. Search for the broken reference, it will either be an asset ID which is a combination UUID and Sub ID, or it will be a reference via relative path.
+
+If you can find the missing reference and it is an Asset ID reference, then you can use the UUID to determine the Source Asset. From here, you can look at the history of this Source Asset in your source control to see what changes may have come in, as well as who authored those changes. You can experiment by rolling back the file locally to see if the missing product asset shows up again.
+
+If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where this thrash is coming from.
 
 ## View Asset Processor Logs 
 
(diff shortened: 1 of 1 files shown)


lmbr-pip (commented):
Thanks for writing this.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
 ---
 
 If you are having issues with **Asset Processor**, use the methods below to debug the issues.
-
-
+
+## Who normally needs to debug Asset Processor issues?
```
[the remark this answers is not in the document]
AMZN-stankowi: I'll get rid of it, I was on the fence on it. It's useful to people maintaining the engine to think about the audience for these issues, but I don't think it needs to be here.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
...
+1. Assign the cube Mesh_A from shapes.fbx to this component.
+1. Save and quit the Editor.
+1. Replace the shapes.fbx file in your project with shapes_2.fbx.
+1. Load the Editor.
+1. Load the prefab you saved earlier.
+1. Notice that the mesh on the entiyt component is still Mesh_A, but now visually looks like the cylinder.
+
+This happens because scene processing tracks the meshes by name, and by swapping the names between the meshes you've replaced which product assets are which. The prefab tracks the outgoing product asset reference via asset ID. When I tested this locally, my Mesh_A azmodel had the Product Asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:270123712, and Mesh_B has the asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:285065766. When I swapped the FBX files, the node within the file named Mesh_A still had that same first ID, but the geometry was a cylinder now, changing what it looked like in the prefab.
+
+Note that in this case, this behavior is 100% intentional. There is no bug in this example, but other times this thrash occurs it may be considered a bug.
```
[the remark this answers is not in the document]
AMZN-stankowi: I updated this, I realized I was still demonstrating a bug but it was a content bug.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
...
+##### Non-deterministic Sub ID generation from builder
+Sometimes the logic for how a builder may not be fully deterministic. For example, if a builder just increments an integer to use as the Sub ID for each product it generates, then a source asset change that results in a change in the products to remove one, will result in the Sub IDs changing for all product assets after the removed product asset.
+
+##### Change in builder logic changes Sub ID generation
+There are times when a builder author needs to make a change to how the builder generates Sub IDs for product assets, resulting in all product assets output from that builder changing Sub IDs the next time they are processed.
+
+#### Solutions
+
+##### Stabilize Sub IDs while authoring a builder
+It is the responsibility of the person authoring a builder to define the rules for how Sub IDs are generated by that builder. When doing so, we recommend putting in effort to stabilize Sub ID generation as much as possible: Think about how the source asset can be modified, and how you can maintain continuity in product asset Sub ID generation based on those changes.
```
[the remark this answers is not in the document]
AMZN-stankowi: I've started a tips section for that.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
...
+
+##### Fix the broken references - update the source asset
+Updating the source asset in a way that it will output a product asset with the same sub ID will fix any broken connections. This works best when the product asset was removed unintentionally, either the content creator was unaware the product asset was in use, or they made a change to the source asset not realizing it would cause this issue.
+
+##### Fix the broken references - update the content referencing the missing product asset
+If the removal of the product asset was intentional, or the source asset cannot safely be updated to restore the product asset, it may be necessary to instead update everything referencing the broken product asset. In this case you'll need to search for all broken references, and for each type or content or code referencing the missing product asset, update it within the relevant system to handle the removal of this product asset or change to the Sub ID.
```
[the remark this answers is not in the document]
AMZN-stankowi: I've added an example. But it really is going to come down to the specific broken reference, every piece of content has its own editor, so to update the reference you'll need to load the specific tool for editing the content that has the broken reference.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
...
+
+#### Debugging
+If content appears missing or broken, try opening the file that references broken content in a text editor or another editor that might show additional details. Search for the broken reference, it will either be an asset ID which is a combination UUID and Sub ID, or it will be a reference via relative path.
+
+If you can find the missing reference and it is an Asset ID reference, then you can use the UUID to determine the Source Asset. From here, you can look at the history of this Source Asset in your source control to see what changes may have come in, as well as who authored those changes. You can experiment by rolling back the file locally to see if the missing product asset shows up again.
+
+If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where this thrash is coming from.
```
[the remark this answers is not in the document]
AMZN-stankowi: I thought about that, but builders are going to be coming from several places: The game team themselves, 3rd parties, or O3DE. Each of these is going to have its own workflow. I'll add a note that for O3DE builders they can cut a ticket.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -43,19 +34,23 @@ Extraneous information is included in the product asset, such as a timestamp. Lo
 In most cases this will require an engineer to make changes to the internal builder logic to stabilize the product output.
 
 #### Debugging
-Your team most likely encounter this issue when generating [Asset Bundles](/packaging/asset-bundler/), which is the primary system it impacts. The Asset Bundling process is generally done by an engineer, who will likely work backward from first finding a product asset in bundles that was not expected to be there, because no changes were done that should have resulted in a new product asset.
+Your team most likely encounter this issue when generating [Asset Bundles](/packaging/asset-bundler/), which is the primary system non-deterministic Product Assets impacts. If a Product Asset shows up in a bundle that you didn't expect to be there because the Source Asset and associated Asset Builder did not change, then debugging tends to start by working backward from there.
```
amzn-mike, line 37: Typo: Your team _will_

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+To help identify the block of logic in the Product Asset that is changing each time it is processed, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
+
+Another debugging option is setting up automation to catch this early. If you have an automated system that processes assets for your project, you can expand this automation to also
```
amzn-mike, line 43: I think this sentence got cut off

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+#### Example
+Here is an example with scene files and prefabs:
+1. Create a new FBX file named shapes_1.fbx with two meshes in it, a cube and a cylinder. Name the cube Mesh_A and the cylinder Mesh_B.
+1. Swap the names of these two meshes, so the cube is now named Mesh_B and the cylinder is named Mesh_A. Save this to a second FBX named shapes_2.fbx.
+1. Place shapes_1.fbx in your project, and rename it to shapes.fbx.
+1. Create a new prefab with an entity in it with a mesh component.
+1. Assign the cube Mesh_A from shapes.fbx to this component.
+1. Save and quit the Editor.
+1. Replace the shapes.fbx file in your project with shapes_2.fbx.
+1. Load the Editor.
```
amzn-mike, line 65: I think the "rename to shapes.fbx" part is missing here

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+1. Load the prefab you saved earlier.
+1. Notice that the mesh on the entiyt component is still Mesh_A, but now visually looks like the cylinder.
```
amzn-mike, line 67: Typo: entiyt

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+This happens because scene processing tracks the meshes by name, and by swapping the names between the meshes you've replaced which product assets are which. The prefab tracks the outgoing product asset reference via asset ID. When I tested this locally, my Mesh_A azmodel had the Product Asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:270123712, and Mesh_B has the asset ID {00431C65-528C-570B-8A52-5D43679BBFAC}:285065766. When I swapped the FBX files, the node within the file named Mesh_A still had that same first ID, but the geometry was a cylinder now, changing what it looked like in the prefab.
```
amzn-mike, line 69: The end of the first sentence is confusing.  I think a comma after "between the meshes" might help (assuming I've understood correctly) but a slight rewording might make it more clear.

content/docs/user-guide/assets/asset-processor/debugging.md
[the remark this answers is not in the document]
amzn-mike: There is a legacy sub-id system that a builder can use in this situation to remap old subids to a new subid system

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+#### Solutions
+
+##### Stabilize Sub IDs while authoring a builder
+It is the responsibility of the person authoring a builder to define the rules for how Sub IDs are generated by that builder. When doing so, we recommend putting in effort to stabilize Sub ID generation as much as possible: Think about how the source asset can be modified, and how you can maintain continuity in product asset Sub ID generation based on those changes.
+
+Tips for stabilizing Sub IDs for Product Assets authoring a builder:
+* A randomly generated Sub ID is usually going to result in the Sub ID changing when not expected, even if the generation is semi-randomized. Randomly generated Sub IDs should be avoided.
+* Often times, the obvious first way to assign Sub ID for Product Assets is based on the ordered index of that information in the Source Asset. Often times this results in problems with Sub IDs changing, because changes in the Source Asset can result in the ordered list used to shuffle. For example if the content creator removes an entry and causes everything else to shuffle down an index.
```
amzn-mike, line 91: Nit: often times repeated.

amzn-mike, line 91: "can result in the ordered list used to shuffle" is confusing.  Perhaps: "can result in the ordered list shuffling"

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+* Hashing the name of an object to use as a reference can work as long as content creators don't expect to rename these objects have have the Sub ID remain stable.
```
amzn-mike, line 92: Typo: have have -> and have

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+#### Debugging
+Debugging this issue is usually focused around discovery of the issue in the first place. Once it's identified as a problem, identifying which builder needs a version change and making that change is usually trivial.
+
+If some team members are experiencing issues with content, but not everyone on the team is having the problem, but that content has not changed recently, this is commonly the cause. Check if the builder for that content has been modified in source control recently, and if so, check if the version was changed. If not, then it's likely that the builder change without a version number change is causing a problem.
+
+The Asset Processor UI lets you right click a Source Asset in the asset tab and re-run all jobs on that asset. If doing so causes a change in Product Asset, it likely means a builder was changed without having the version updated. If you want to debug this back and force, it's recommended that you back up the old Product Asset(s) before reprocessing the Source Asset.
```
amzn-mike, line 159: Typo: force -> forth

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+### Logic in Asset Builder persists between building assets
+
+#### Situation
+In some cases, due to information persisting during asset processing, you may get different Product Asset output from jobs based on the order Source Assets are processed. This is usually due to a bug within the Asset Builder that processed this content.
+
+The Asset Processor creates one or more Asset Builder executable based on configuration settings, and uses these to process assets, sending jobs to these Asset Builder executables to run with the individual Asset Builders. There is no system within the Asset Builder Executable to completely shut down and clear all Asset Builders, so it is possible for an Asset Builder to persist information across multiple sessions of processing jobs for a single Asset Builder Executable.
```
amzn-mike, line 166: Typo: executable -> executables
Nit: creates -> starts/runs

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+#### Effect
+Stale data or logic running in an Asset Builder in an Asset Builder Executable can result in an Asset processing incorrectly.
```
amzn-mike, line 169: This reads like a typo but I understand it... Not sure if there's some way to reword it so it doesn't look like you've accidentally doubled "in an Asset Builder".  Maybe just drop the 2nd part, it should be understood that Asset Builders run in AssetBuilder.exe by this point?

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+#### Cause
+The cause for this issue is when information is cached during one step in processing to make it easier to access at a later step.
+
+#### Example
+This example covers a situation we encountered this problem:
```
amzn-mike, line 175: situation _where_ we

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+>	Element with class ID '{4C19E257-F929-524E-80E3-C910C5F3E2D9}' is not registered with the serializer!
+```
+
+This warning occured when a prefab file referenced a class that was no longer registered with the serializer. This is an error because this prefab will no longer behave the same way it behaved when it was authored, because there is no longer a C++ class that matches this serialization information. Because this is likely not intentional and could have side effects that effect the functionality of this prefab, this is an error instead of a warning. Because the prefab can still be processed in some other ways, it is an error instead of a failure. In this case a broken Product Asset is usually desired over no Product Asset at all, because it gives more options for debugging and correcting the situation.
+
+#### Example - Failure
+```
+>	Failed to import Asset Importer Scene. Error returned: FBX-Parser unexpected end of file
+```
+This failure occured because the file I attempted to process was a blank file with the FBX extension. There was no information for the Scene Builder to even attempt to create a Product Asset from, so it failed instead and output nothing.
```
amzn-mike, line 231: Should we avoid writing in first person for these docs?

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+Usually clearing this will involve a change to the Source Asset. In some cases it may make more sense to change the Asset Builder to handle this data more correctly. If you're a content creator and you don't see a path you like in addressing the issue, talk to someone on your engineering team to see if the Asset Builder can be updated to better handle the Source Asset.
+
+#### Debugging
+The first step in debugging asset job warnings, errors, and failures is the [jobs tab](interface/#jobs) of the Asset Processor UI, or the log output of Asset Processor Batch. The logging from processing the job generally give guidance on why this situation is a problem and in many cases covers steps to take to correct it.
+
+The next debugging option is checking the documentation for that specific builder may provide guidance on creating assets that process cleanly. For example, if you are encountering an issue with FBX files and aren't sure what steps to take, the [3D Scene Format Support documentation](/assets/scene-settings/scene-format-support/) may have sone information to help address the issue.
```
amzn-mike, line 241: specific builder may -> specific builder, which may

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,256 @@ weight: 500
...
+
+To help with that, it may be useful to [more directly debug the Asset Builder](#DebugAssetBuilders).
+
+### Slow Asset Processing Times
+Long asset processing times can be disruptive to a team's ability to iterate quickly on a project.
+
+#### Common Causes
+* Source asset quantity - A project with a lot of content will take a longer time to process.
+* Large or difficult to process source assets - In some cases, a single source asset may be an outlier, having a much longer processing time than most other content.
+* Deep job dependency web - A a large job dependency web can result in a change to one file causing many other files to need to reprocess. See (this documentation for more details)[/assets/pipeline/asset-dependencies-and-identifiers/].
```
amzn-mike, line 253: Typo: a a

content/docs/user-guide/assets/asset-processor/debugging.md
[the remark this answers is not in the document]
AMZN-stankowi: I added a small note on legacy sub IDs. When I update the Asset Builder documentation, I'll include more details, and update the debugging page to point at that.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
 ---
 
 If you are having issues with **Asset Processor**, use the methods below to debug the issues.
```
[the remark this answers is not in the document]
AMZN-stankowi: I updated the text to describe the page better.

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: I've reworded this

amzn-mike (approved):

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -97,6 +347,7 @@ Use this tool to automatically attach the debugger to spawned child processes.
 1. In Visual Studio, start `AssetProcessor.exe`. Breakpoints in Asset Builders work as normal.
 
 ## Debug Asset Builders from Asset Processor 
+<a name="DebugAssetBuilders"></a>
```
sptramer, line 350:
```suggestion
<a name="DebugAssetBuilders"/>
```

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+#### Debugging
+The first step in debugging asset job warnings, errors, and failures is the [jobs tab](interface/#jobs) of the Asset Processor UI, or the log output of Asset Processor Batch. The logging from processing the job generally give guidance on why this situation is a problem and in many cases covers steps to take to correct it.
+
+The next debugging option is checking the documentation for that specific builder, which may provide guidance on creating assets that process cleanly. For example, if you are encountering an issue with FBX files and aren't sure what steps to take, the [3D Scene Format Support documentation](/assets/scene-settings/scene-format-support/) may have sone information to help address the issue.
+
+Another option would be to examine the source code for the builder itself, especially around the locations generating those messages. Step through why the builder is having issue with the data, and see if this either leads to a way to change the Source Asset to clear the issue, or a path to changing the builder itself to better handle the Source Asset.
+
+To help with that, it may be useful to [more directly debug the Asset Builder](#DebugAssetBuilders).
```
sptramer, line 248:
````suggestion
## Fixing warnings, errors, and failures

After an asset finishes processing, if the job completed it may still have a number of warnings or errors you may wish to address. In other cases, the job may not finish and may instead fail.

You can see information about viewing Asset Job status in the Asset Processor [here](interface/#jobs).

### Effect

A warning or error in an Asset Job is the Asset Builder author's way of informing content authors that they encountered an issue processing your asset, and may not have generated the Product Asset the content author expected.

A failure is an Asset Builder author's way of informing content authors that they were unable to process the Source Asset at all and cannot put out product assets.

The builder author decides on a case by case basis what a warning, error, or failure is. There is guidance [here](/docs/user-guide/assets/asset-processor/interface/#jobs), with suggested patterns to follow.

### Example - Warning

```
>	The name of the node 'Material.003' was invalid or conflicting and was updated to 'Material_003'.
```

This warning occurred when an FBX file contained a material with the illegal character `.` in its name. This is a warning because the system was able to safely substitute the character for another, legal character. No action needs to be taken, but if the content creator wanted to clear this warning, they would rename the material in the FBX file. In this example, if you did decide to rename the material to clear the warning, be careful you don't introduce another problem by causing the [Sub ID of the Product Asset to change](#ProductAssetSubId).

### Example - Error

```
>	Element with class ID '{4C19E257-F929-524E-80E3-C910C5F3E2D9}' is not registered with the serializer!
```

This warning occurred when a prefab file referenced a class that was no longer registered with the serializer. This is an error because this prefab will no longer behave the same way it behaved when it was authored, because there is no longer a C++ class that matches this serialization information. Because this is likely not intentional and could have side effects that effect the functionality of this prefab, this is an error instead of a warning. Because the prefab can still be processed in some other ways, it is an error instead of a failure. In this case a broken Product Asset is usually desired over no Product Asset at all, because it gives more options for debugging and correcting the situation.

### Example - Failure

```
>	Failed to import Asset Importer Scene. Error returned: FBX-Parser unexpected end of file
```

This failure occurred because the file processed was a blank file with the FBX extension. There was no information for the Scene Builder to even attempt to create a Product Asset from, so it failed instead and output nothing.

### Solutions

Most Warning, Error, and Failure messages will include enough information to help guide the steps to take in addressing the problem. If this is not enough information, the next step would be to look at other log messages for that job, to see if they contain relevant information.

Usually clearing this will involve a change to the Source Asset. In some cases it may make more sense to change the Asset Builder to handle this data more correctly. If you're a content creator and you don't see a path you like in addressing the issue, talk to someone on your engineering team to see if the Asset Builder can be updated to better handle the Source Asset.

### Debugging

The first step in debugging asset job warnings, errors, and failures is the [jobs tab](/docs/user-guide/assets/asset-processor/interface/#jobs) of the Asset Processor UI, or the log output of Asset Processor Batch. The logging from processing the job generally give guidance on why this situation is a problem and in many cases covers steps to take to correct it.

The next debugging option is checking the documentation for that specific builder, which may provide guidance on creating assets that process cleanly. For example, if you are encountering an issue with FBX files and aren't sure what steps to take, the [3D Scene Format Support documentation](/docs/user-guide/assets/scene-settings/scene-format-support/) may have some information to help address the issue.

Another option would be to examine the source code for the builder itself, especially around the locations generating those messages. Step through why the builder is having issue with the data, and see if this either leads to a way to change the Source Asset to clear the issue, or a path to changing the builder itself to better handle the Source Asset.

To help with that, it may be useful to [more directly debug the Asset Builder](#DebugAssetBuilders).

````

File: content/docs/user-guide/assets/asset-processor/debugging.md [hunk 23]
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+Long asset processing times can be disruptive to a team's ability to iterate quickly on a project.
+
+#### Common Causes
+* Source asset quantity - A project with a lot of content will take a longer time to process.
+* Large or difficult to process source assets - In some cases, a single source asset may be an outlier, having a much longer processing time than most other content.
+* Deep job dependency web - A large job dependency web can result in a change to one file causing many other files to need to reprocess. See (this documentation for more details)[/assets/pipeline/asset-dependencies-and-identifiers/].
+
+#### Solutions
+* Removing unused content from your project can save on overall asset processing time.
+* Targeted performance improvements at the Asset Builders that are taking the most overall time in asset processing can help.
```
sptramer, line 260:
```suggestion
## Slow Asset Processing Times

Long asset processing times can be disruptive to a team's ability to iterate quickly on a project.

### Common Causes

* Source asset quantity - A project with a lot of content will take a longer time to process.
* Large or difficult to process source assets - In some cases, a single source asset may be an outlier, having a much longer processing time than most other content.
* Deep job dependency web - A large job dependency web can result in a change to one file causing many other files to need to reprocess. Read the topic on [asset dependencies and identifiers](/docs/user-guide/assets/pipeline/asset-dependencies-and-identifiers/) for more information.

### Solutions

* Removing unused content from your project can save on overall asset processing time.
* Targeted performance improvements at the Asset Builders that are taking the most overall time in asset processing can help.
```

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+Attaching Visual Studio to a running Asset Builder can also help with debugging. Following [these instruction](#DebugAssetBuilders) will run Asset Processor with a single builder, so you can process multiple assets in order with Visual Studio connected to the Asset Builder Executable. At that point you can try manually re-processing assets using the right click menu of the Asset Tab, and track what data is persisting across multiple jobs on the same Asset Builder on the same Asset Builder Executable.
```
sptramer, line 201:
```suggestion
## Logic in Asset Builder persists between building assets

In some cases, due to information persisting during asset processing, you may get different Product Asset output from jobs based on the order Source Assets are processed. This is usually due to a bug within the Asset Builder that processed this content.

The Asset Processor starts one or more Asset Builder executables based on configuration settings, and uses these to process assets, sending jobs to these Asset Builder executables to run with the individual Asset Builders. There is no system within the Asset Builder Executable to completely shut down and clear all Asset Builders, so it is possible for an Asset Builder to persist information across multiple sessions of processing jobs for a single Asset Builder Executable.

### Effect

Stale data or logic in the Asset Builder Executable can result in an Asset processing incorrectly.

### Cause

The cause for this issue is when information is cached during one step in processing to make it easier to access at a later step.

### Example

This example covers a situation where this problem was encountered:

* The Scene Builder processes scene files, such as FBX files, into a collection of product assets.
* Each scene file has a scene manifest that contains additional processing rules.
* Within the scene manifest, a Python script file can be marked to run during scene processing.
* On initial startup, the Scene Builder within each Asset Builder Executable would start with no set Python script.
* On encountering a scene file that had a scene manifest with a Python script to run, the Scene Builder was storing this Python script to use later.
* The Scene Builder unintentionally stored this data in a way that persisted across jobs.
* For this example, FBX file `NoPython.fbx` does not have a Python script set to run in its Scene Manifest, but FBX `WithPython.fbx` does have a Python script set to run in its Scene Manifest.
* If both jobs end up running on the same Asset Builder Executable, then if `WithPython.fbx` is processed by the Scene Builder on that Asset Builder before `NoPython.fbx`, the Python script to run was persisted in a way that it was run on `NoPython.fbx`.
* The end result was `NoPython.fbx` was producing different output if it ran after `WithPython.fbx` on the same Asset Builder Executable as `WithPython.fbx` ran on over any other situation: Running before `WithPython.fbx`, running on a different Asset Builder Executable.

### Solutions

The primary solution to this problem is to update the builder to not cache information in a way that it persists across jobs.

Automated tests can help prevent this from cropping up. You can see a test that covers the example case above [here](https://github.com/o3de/o3de/blob/development/AutomatedTesting/Gem/PythonTests/assetpipeline/fbx_tests/pythonassetbuildertests.py#L50). This test is annotated with how it's setup and why.

### Debugging

This issue is difficult to identify because it requires a very specific setup, as explained previously. This issue was found by working backward from intermittent issues on automated asset processing jobs on the O3DE Jenkins servers.

If your team is seeing situations where some people encounter different asset processing results than other team members, and you've already ruled out [a builder change without a version change](#NoVersionChange) as the source of the issue, then keep this in mind. Examine asset processing logs for the assets showing the issue, and check the asset processing order for cases where the issue occurs, compared against cases it does not occur. Specifically look for what other jobs ran on the same Asset Builder Executable, using the same Asset Builder on that executable.

Asset Processor, both the command line and graphical interfaces, support a command line flag that can help with investigating. The flag `--sortJobsByDBSourceName` will stabilize the order that jobs are run in. Using this while debugging will let you test different job processing orders by renaming assets to control the order they run. The Asset Processor also allows for regset values to be controlled via command line, using the `--regset` flag with the setting to set. Specifically, setting the regset value `/Amazon/AssetProcessor/Settings/Jobs/maxJobs` to `1` will restrict the Asset Processor to only launch a single Asset Builder executable. Note that if you have many assets to process, this will result in a long asset processing time, so it's recommended that you set this after you've processed all jobs in a previous Asset Processor run with more max jobs. Once you are running Asset Processor with one max job, all assets will process in that session on the same builder, letting you specifically process assets in the order of your choosing.

Attaching Visual Studio to a running Asset Builder can also help with debugging. Following [these instructions](#DebugAssetBuilders) will run Asset Processor with a single builder, so you can process multiple assets in order with Visual Studio connected to the Asset Builder Executable. At that point you can try manually re-processing assets using the right click menu of the Asset Tab, and track what data is persisting across multiple jobs on the same Asset Builder on the same Asset Builder Executable.
```

I think the example provided here could be greatly simplified, but without the full technical details, I'm unsure how. It looks like a test artifact was being cached and inappropriately loaded for another test, depending on the order of the test runs?

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+#### Debugging
+Debugging this issue is usually focused around discovery of the issue in the first place. Once it's identified as a problem, identifying which builder needs a version change and making that change is usually trivial.
+
+If some team members are experiencing issues with content, but not everyone on the team is having the problem, but that content has not changed recently, this is commonly the cause. Check if the builder for that content has been modified in source control recently, and if so, check if the version was changed. If not, then it's likely that the builder change without a version number change is causing a problem.
+
+The Asset Processor UI lets you right click a Source Asset in the asset tab and re-run all jobs on that asset. If doing so causes a change in Product Asset, it likely means a builder was changed without having the version updated. If you want to debug this back and forth, it's recommended that you back up the old Product Asset(s) before reprocessing the Source Asset.
```
sptramer, line 162:
```suggestion
## Logic change in builder with no version change
<a name="NoVersionChange"/>

An engineer makes changes to the logic of an asset builder, but the engineer does not change the fingerprint or version of the builder.

### Common Issues

Source Assets processed by this builder will not be automatically reprocessed with the latest changes.

If an asset would fail to process due to the builder change, the engineer making the change might not notice until the builder change is pushed to the rest of the team, causing other people to be disrupted by the failing asset.

If other content or logic requires the changes from processing the asset, other team members may run into problems using the Product Assets output by the older version of the builder with the other updated logic.

### Immediate Solution

Change the version number of the builder reported in that builder's `AssetBuilderSDK::AssetBuilderDesc`.

### Long Term Solutions

Create automated jobs that process all assets cleanly at some interval, to catch issues like this.

One configuration that would catch issues would be:

* Iterative Job - Processes assets after a code rebuild.
* Clean Job - Cleans and processes all assets after a code build.
* Compare the hashes of each Product Asset between builds and runs of these jobs. If any assets end up different in two back to back Clean Jobs but did not process in the Iterative Job, this usually means an asset builder was modified without changing its version.

### Debugging

Debugging this issue is usually focused around discovery of the issue in the first place. Once it's identified as a problem, identifying which builder needs a version change and making that change is usually trivial.

If some team members are experiencing issues with content, but not everyone on the team is having the problem, but that content has not changed recently, this is commonly the cause. Check if the builder for that content has been modified in source control recently, and if so, check if the version was changed. If not, then it's likely that the builder change without a version number change is causing a problem.

The Asset Processor UI lets you right click a Source Asset in the asset tab and re-run all jobs on that asset. If doing so causes a change in Product Asset, it likely means a builder was changed without having the version updated. If you want to debug this back and forth, it's recommended that you back up the old Product Asset(s) before reprocessing the Source Asset.
```

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+#### Debugging
+If content appears missing or broken, try opening the file that references broken content in a text editor or another editor that might show additional details. Search for the broken reference, it will either be an asset ID which is a combination UUID and Sub ID, or it will be a reference via relative path.
+
+If you can find the missing reference and it is an Asset ID reference, then you can use the UUID to determine the Source Asset. From here, you can look at the history of this Source Asset in your source control to see what changes may have come in, as well as who authored those changes. You can experiment by rolling back the file locally to see if the missing product asset shows up again.
+
+If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where the change in Sub IDs is coming from.
+
+If the builder is an O3DE builder, and not one that your team has created, you can create a ticket [here](https://github.com/o3de/o3de/issues) to get this looked at.
```
sptramer, line 131:
```suggestion
## Same Product Asset generated with different Sub ID
<a name="ProductAssetSubId"/>

The Sub ID for a Product Asset changes when it is not expected to change. In the worst case, the Sub ID is changing when no changes have been made to the Source Asset or builder that created the Product Asset. In other cases, a change to the Source Asset results in a Product Asset that has a different Sub ID than it did previously.

### Effect

Assets are mostly referenced by asset ID, the UUID of the source asset with the Sub ID for that product asset. If this is not consistent and the Sub ID changes for a Product Asset when it should not change, then external references to this asset can end up incorrect, pointing at a different asset after a rebuild or pointing at no asset if no other products use that Sub ID.

### Example

Here is an example with scene files and prefabs:

1. Create a new FBX file named `shapes_1.fbx` with two meshes in it, a cube and a cylinder. Name the cube `Mesh_A` and the cylinder `Mesh_B`.
1. Swap the names of these two meshes, so the cube is now named `Mesh_B` and the cylinder is named `Mesh_A`. Save this to a second FBX named `shapes_2.fbx`.
1. Place `shapes_1.fbx` in your project, and rename it to `shapes.fbx`.
1. Create a new prefab with an entity in it with a mesh component.
1. Assign the cube `Mesh_A` from `shapes.fbx` to this component.
1. Save and quit the Editor.
1. Replace the `shapes.fbx` file in your project with the `shapes_2.fbx` file created earlier.
1. Load the Editor.
1. Load the prefab you saved earlier.
1. Notice that the mesh on the entity component is still `Mesh_A`, but now visually looks like the cylinder.

This happens because scene processing generates Sub IDs based on the node's name. The prefab was tracking the outgoing asset reference via this ID, so the node name is functionally what is used to track references to individual Product Assets generated from the source FBX.

Note that in this case, this behavior is 100% intentional. There is no code issue in this example. In this case, if this change was not intended by the FBX author, it is a content bug. The following sections will cover how to identify and correct issues like this.

### Common Causes

#### Change to source asset causes Sub IDs to shift

The most common place Sub ID changes occurs is when the Sub ID generation is based on data within the source asset, and that data can change. For example, if a node name is hashed to generate the Sub ID for a product asset, then a content creator renaming that node in the source asset will cause the associated product asset to generate a new Sub ID.

#### Non-deterministic Sub ID generation from builder

Sometimes the logic for how a builder may not be fully deterministic. For example, if a builder just increments an integer to use as the Sub ID for each product it generates, then a source asset change that results in a change in the products to remove one, will result in the Sub IDs changing for all product assets after the removed product asset.

#### Change in builder logic changes Sub ID generation

There are times when a builder author needs to make a change to how the builder generates Sub IDs for product assets, resulting in all product assets output from that builder changing Sub IDs the next time they are processed.

### Solutions

#### Stabilize Sub IDs while authoring a builder

It is the responsibility of the person authoring a builder to define the rules for how Sub IDs are generated by that builder. When doing so, we recommend putting in effort to stabilize Sub ID generation as much as possible: Think about how the source asset can be modified, and how you can maintain continuity in product asset Sub ID generation based on those changes.

The Asset Pipeline supports a Legacy Sub ID that can be used to remap old sub IDs to new sub IDs, when a builder author needs to change them.

Tips for stabilizing Sub IDs for Product Assets authoring a builder:

* A randomly generated Sub ID is usually going to result in the Sub ID changing when not expected, even if the generation is semi-randomized. Randomly generated Sub IDs should be avoided.
* Sometimes builder authors will assign Sub ID for Product Assets based on the ordered index of that information in the Source Asset. This can result in unstable Sub IDs, because changes in the Source Asset can result in the ordered list shuffling.
   * For example if the content creator removes an entry and causes everything else to shuffle down an index.
* Hashing the name of an object to use as a reference can work as long as content creators don't expect to rename these objects to have the Sub ID remain stable.
   * Hashing the name of an object to use as the Sub ID can also be beneficial to content creators in that it allows them to replace content by creating new data with the same name to generate a Product Asset with the same ID.
* Sometimes there may be a stable internal ID you can use, that content creators do not get direct control over. For example, there may be an internal node structure that generates unique internal IDs for nodes that content creators cannot change. This may be more stable in some ways to use to generate Sub IDs, but consider your content creator's workflows before committing to using this for Sub ID generation. There may be times a content creator specifically wants to make a new thing to replace something, and have the new Product Asset seamlessly be used everywhere the old one was. See the previous point about using object names for an example.


#### Learning how Sub ID generation works for builders that process content you author

As a content creator, learning how Sub IDs are generated for Product Assets created from the Source Assets you work on can prevent these problems in the first place.

When a Sub ID changes for a Product Asset, it will cause all references to that Product Asset to break, because the Asset ID will no longer resolve to that asset.

During initial content creation, especially for scene files like FBX, thinking through how you want that content referenced elsewhere can help keep the Sub ID for the Product Assets stable.

If you need to make a change to the Source Asset, keep in mind how your changes will impact the Sub IDs of the Product Assets. If you do need to change the Product Assets, a good pattern to follow is:

1. Create a stub of the new Product Asset, or leave a stub for the old Product Asset.
1. Find all references to the old Product Asset.
1. Update those as necessary, working with other members on your team to do so.
1. Once you're sure nothing is left referencing the old Product Asset, then make the final change to remove it from the Source Asset.

#### Fix the broken references - update the source asset

Updating the source asset in a way that it will output a product asset with the same sub ID will fix any broken connections. This works best when the product asset was removed unintentionally, either the content creator was unaware the product asset was in use, or they made a change to the source asset not realizing it would cause this issue.

#### Fix the broken references - update the content referencing the missing product asset

If the removal of the product asset was intentional, or the source asset cannot safely be updated to restore the product asset, it may be necessary to instead update everything referencing the broken product asset. In this case you'll need to search for all broken references, and for each type or content or code referencing the missing product asset, update it within the relevant system to handle the removal of this product asset or change to the Sub ID.

For example, if an FBX file was updated to no longer generate a Product Asset, and an O3DE Prefab referenced this product asset via a mesh component on an entity, you can use the prefab editing workflow to fix this.

1. Load the prefab to be edited.
1. Find the component with the broken reference.
1. Either change it to reference alternate content, or remove the component/entity.

### Debugging

If content appears missing or broken, try opening the file that references broken content in a text editor or another editor that might show additional details. Search for the broken reference, it will either be an asset ID which is a combination UUID and Sub ID, or it will be a reference via relative path.

If you can find the missing reference and it is an Asset ID reference, then you can use the UUID to determine the Source Asset. From here, you can look at the history of this Source Asset in your source control to see what changes may have come in, as well as who authored those changes. You can experiment by rolling back the file locally to see if the missing product asset shows up again.

If you believe the issue is within the builder itself, and is not a content problem, then you can try [debugging the Asset Builder](#DebugAssetBuilders) to investigate how the product Sub IDs are generated and where the change in Sub IDs is coming from.

If the builder is an O3DE builder, and not one that your team has created, you can create a ticket [here](https://github.com/o3de/o3de/issues) to get this looked at.
```

sptramer (changes requested):
Some formatting and broken link changes, mostly looks good!

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -7,8 +7,105 @@ toc: true
...
+* After changing a source asset, it takes too long for assets to finish processing.
+* The product assets don't behave like expected. They may be missing entirely, missing information, have incorrect information.
+* Finding and attempting to fix a warning, error, or failure that occured during asset processing for a particular job.
+* Asset packaging isn't working as expected, content is missing or content is being included in bundles that was not expected to be there.
+* Assistant while working on a builder to add a new feature, fix a bug, or create a new builder.
+* Different members of a team getting different results from asset processing, impacting the ability for the team to cooperate.
+
+## What are common Asset Processor related problems, and paths to address those problems?
+
+### Product Asset Thrash - Non Deterministic Product Assets
```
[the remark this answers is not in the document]
AMZN-stankowi: I've removed the word thrash from this document. I've tried to be more clear on non-deterministic, I couldn't think of an easier to understand alternative though.

content/docs/user-guide/assets/asset-processor/debugging.md
willihay: Netlify-style anchor links work better. Using HTML links will cause the page to scroll too far.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+To help identify the block of logic in the Product Asset that is changing each time it is processed, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
+
+Another debugging option is setting up automation to catch this early. If you have an automated system that processes assets for your project, you can expand this automation to also compare iterative asset processing to a clean cache asset process and the previous day's assets. Doing this will let you compare the hashes of Product Assets, and find any where the hash has changed when it should not have.
+
+### Same Product Asset generated with different Sub ID
+<a name="ProductAssetSubId"></a>
```
willihay, line 46: Please change all HTML `<a>`  links in this PR to Netlify links. Using Netlify links on the same line as the header will make scrolling to this link work properly.
```suggestion
### Same Product Asset generated with different Sub ID {#ProductAssetSubId}
```

willihay (changes requested):

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: > I think the example provided here could be greatly simplified, but without the full technical details, I'm unsure how. It looks like a test artifact was being cached and inappropriately loaded for another test, depending on the order of the test runs?

It's a complex situation that I'm struggling to figure out how to describe in a simple way. The Asset Builder Executables run whatever code you give them. So if you're not careful as an Asset Builder author, you can accidentally use information from an asset processed earlier to change how an asset is processed later.

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+## What kinds of problems can be encountered within Asset Processing?
+
+* After changing a source asset, it takes too long for assets to finish processing.
+* The product assets don't behave like expected. They may be missing entirely, missing information, have incorrect information.
+* Finding and attempting to fix a warning, error, or failure that occured during asset processing for a particular job.
+* Asset packaging isn't working as expected, content is missing or content is being included in bundles that was not expected to be there.
+* Assistant while working on a builder to add a new feature, fix a bug, or create a new builder.
+* Different members of a team getting different results from asset processing, impacting the ability for the team to cooperate.
+
+## What are common Asset Processor related problems, and paths to address those problems?
```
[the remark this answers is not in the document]
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+### Non Deterministic Product Assets - Different Product Assets with No Source Changes
```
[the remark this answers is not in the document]
AMZN-stankowi: Done, however will this be clear enough? Or is there an alternate term we can use to "non-deterministic"?

content/docs/user-guide/assets/asset-processor/debugging.md
```diff
@@ -6,9 +6,259 @@ weight: 500
...
+
+To help identify the block of logic in the Product Asset that is changing each time it is processed, you may be able to use a standard file diffing tool using two generated product assets. You can quickly re-generate the product asset by right clicking it in the Asset Processor's asset tab and selecting reprocess source asset.
+
+Another debugging option is setting up automation to catch this early. If you have an automated system that processes assets for your project, you can expand this automation to also compare iterative asset processing to a clean cache asset process and the previous day's assets. Doing this will let you compare the hashes of Product Assets, and find any where the hash has changed when it should not have.
```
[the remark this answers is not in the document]
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: Done

content/docs/user-guide/assets/asset-processor/debugging.md
AMZN-stankowi: Changed all of them.

content/docs/user-guide/assets/asset-processor/debugging.md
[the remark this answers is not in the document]
sptramer: Volatile? That sounds too strong, but I do wonder if there's a better term for it. If we think of one later the doc can get updated.

content/docs/user-guide/assets/asset-processor/debugging.md
sptramer: Will take an issue that this example could either be simplified (to be in-line) or expanded (as a separate example of conflicts, with appropriate resources available.)

sptramer (approved):

PR #1530, opened 2022-04-06
updated a few screenshot

amzn-rhhong:
Signed-off-by: rhhong <rhhong@amazon.com>

Update a few screenshot that reflect the change from opengl renderviewport to atom viewport in animation editor.

## Change summary
Screenshot 2022-04-05 203643
Screenshot 2022-04-05 203721
Screenshot 2022-04-05 203733
Screenshot 2022-04-05 203745


hultonha (approved):
This looks great to me! Thanks for jumping on this @amzn-rhhong! Be great to have @micronAMZN review and would be worth sanity checking about the LFS stuf too (@AMZN-alexpete would be the best person to verify). Otherwise looks great, thanks! 🙂

galibzon (approved):

micronAMZN (changes requested):
Due to limitations on the repo size, we have a guideline that images should not exceed 512KB with a hard limit of 1MB. Though there are still some legacy gifs in the docs, we can't accept new gifs. Legacy gifs are being removed or converted to pngs as the topics containing the gifs are updated. We don't have a solution for including animated images in docs at this time. Please see the submission guidelines for images: https://www.o3de.org/docs/contributing/to-docs/style-guide/media/

amzn-rhhong:
> 

@micronAMZN huh. I think replacing the animated gif with a png (or series of pngs) is a downgrade. I already pretty careful on the gif sizes (30fps, reduced frames, etc..) but I cannot get their size down to 512k. It's more close to ~5mb.
If that is the case, I'd rather keep the original gifs until we have a better solution to support large gif/videos.

chanmosq:
@amzn-rhhong We don't currently have resources to support large files in docs. Another option would be to upload a short video on Youtube and embed it, instead of a GIF. We might be able to do it through a private, view-only-with-link through the O3DE channel? (@Ulrick28 might have more info on this.)

If you want to try this route, embed the video by adding this shortcode in the `.md` file:
`{{< youtube-width id="CQmjAxr7LZs" title="What is O3DE?" width="100%" >}}`
- `id` is the last bit of the URL: `https://www.youtube.com/watch?v=`**`CQmjAxr7LZs`**
- You can also set `autoplay="true"` but that might impact page load speed.

micronAMZN (approved):
Approving, we will find a resolution for animated images in the coming weeks as the animation docs are updated.

amzn-rhhong:
thanks for the suggestion @sptramer @micronAMZN 
I like the idea to put a link to a short youtube click. We would probably need to set up some official procedure like what's the length / size /format of the video and where to submit them. I can re-record them using the guideline. Also I imagine we should all put it into an o3de official youtube channel or something

micronAMZN:
@chanmosq suggested trying mp4 again. I am working on a shortcode solution for this. There's no need to do anything with this PR just yet. I will make a new commit to this PR with the fix. Should be within the next day.

sptramer:
Placing PR on hold until Mike's work comes in. Thanks Mike!

micronAMZN:
This PR will depend on #1539

micronAMZN:
I was unable to update this PR directly because it's not on a fork. I have updated this submission to use MP4s in #1567. Closing this PR.

PR #1575, opened 2022-04-21
Updated image-width shortcode to use named parameters.

micronAMZN:
Ordered parameters are still supported for backwards compatibility with prior documentation.

Signed-off-by: Mike Cronin <58789750+micronAMZN@users.noreply.github.com>

### DIFF
--- a/content/docs/contributing/to-docs/style-guide/media.md
+++ b/content/docs/contributing/to-docs/style-guide/media.md
@@ -26,21 +26,21 @@ Image output:
 
 The `image-width` shortcode adds an image with alternate text and restricts the image's width. The `image-width` shortcode can ensure image sizes are consistent within a topic, and that large images don't scale overly large in wide browser windows. Use this method when adding `.svg` diagrams, very large images, and in topics with multiple images where consistent image widths are desirable.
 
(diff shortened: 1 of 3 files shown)


chanmosq:
I was wondering what's the reason for having an even number of pixels in both width and height?

micronAMZN:
> I was wondering what's the reason for having an even number of pixels in both width and height?

When writing the game objects tutorials, I noticed that if an image or video has an odd number of pixels in either width or height some image/video editing tools blur the image a bit when it gets compressed. Particularly noticeable with MP4 videos. It's doesn't make interface screenshots unreadable, but they do sometimes look significantly worse than images with even pixel counts. Images and movies with an even pixel counts are really crisp when displayed at full res. I assume this has something to do with the sampling methods some image and video editing applications use. I suppose that when an image has an odd number of pixels, the sampler might sub-sample pixels instead of landing on pixel edges. I don't think we need to do an audit of every image over this. We can replace particularly bad movies and screenshots as we encounter them.

chanmosq:
Huh, interesting observation! Glad you caught that.

chanmosq (approved):

PR #1597, opened 2022-05-03
Update UI Canvas on mesh feature

michabr:
Signed-off-by: abrmich <abrmich@amazon.com>

The "UI Canvas in 3D" implementation was updated to support Atom renderer. These doc changes reflect that.


FiniteStateGit:
Hello @michabr,

I was unable to find any information on attachment image assets here: https://www.o3de.org/docs/user-guide/assets/asset-types/.   There are a few example assets of the type in the Atom gem, but it is unclear if the ClassData of these examples are appropriate for this use case.  I was also unable to test the process, running the level with the canvas loaded would CTD with the error 'Graphics card was removed.'  Were you able to demo this and can you elaborate on the use of attachment image assets?

michabr:
> Hello @michabr,
> 
> I was unable to find any information on attachment image assets here: https://www.o3de.org/docs/user-guide/assets/asset-types/. There are a few example assets of the type in the Atom gem, but it is unclear if the ClassData of these examples are appropriate for this use case. I was also unable to test the process, running the level with the canvas loaded would CTD with the error 'Graphics card was removed.' Were you able to demo this and can you elaborate on the use of attachment image assets?

Hi, it looks like attachment image assets have not been documented yet. These are assets with the .attimage extension. They represent render targets that are created and drawn to at runtime. They are like textures that have runtime specific content. In this case, a UI canvas is drawn onto a render target which is used as a texture in a material. I've updated the PR to add a little more info on the subject.

FiniteStateGit:
That info you added was enough:
canvasin3dworld

This should resolve https://github.com/o3de/o3de/issues/8691.  Thanks for updating the topic, I'll add my review!

FiniteStateGit:
@michabr How would you set up an entity that shares an interactive canvas but not the canvas' state with another entity?  

Does one entity load the canvas with the Canvas Asset Ref component, whereas other entities use a Canvas Proxy Ref component? Does each entity load a copy of the canvas with the Canvas Asset Ref component?

A canvas will point to a single attachment image asset.  It seems like a second copy of the canvas would be needed to render to a different asset.  Do two entities use the same attachment image asset (as texture and Render Target Override), or are separate assets needed for every instance of canvas sharing?

I was able to create an interactive canvas and share it among different entities, but I couldn't get Render Target Overrides working.

michabr:
> I was able to create an interactive canvas and share it among different entities, but I couldn't get Render Target Overrides working.

I think I found an issue with the render target overrides. I will get back to you on this once I resolve it.

michabr:
> > I was able to create an interactive canvas and share it among different entities, but I couldn't get Render Target Overrides working.
> 
> I think I found an issue with the render target overrides. I will get back to you on this once I resolve it.

PR to fix overrides: https://github.com/o3de/o3de/pull/9396

FiniteStateGit (approved):
Looks good, and thanks for hammering out the details with me.

FiniteStateGit:
Will fast follow with a formatting pass to the UI component references, issue filed https://github.com/o3de/o3de.org/issues/1615.

willihay (approved):

willihay:
@michabr The conflict in canvas-properties.md must be resolved before this PR can be merged. Thanks!

PR #1627, opened 2022-05-12
Use tabs for dependent properties

amzn-tommy:
## Change summary
The PhysX Collider component has a dropdown for 'Shape', with different configurations available depending on which shape is chose. I was looking for an example of how do document similar properties for the mesh component, and the PhsyX Collider component is the given example in the component reference doc. However, I believe this case is best served with tabs, which is what I used for the mesh component. This PR updates the PhysX Collider component to also use tabs, and it updates the component reference doc to give an example of the Light Component which does not use tabs.

This is the difference between what existed previously for the PhysX Collider component, and what it looks like with tabs:
Before
physx_collider_before
After, with tabs
physx_collider_after_withtabs

### Submission Checklist:

* [ x] **Descriptive active voice** - Do descriptive sentences have a clear *subject* and *action verb*?
* [x ] **Answer the question at hand** - Does the documentation answer a *what*, *why*, *how*, or *where* type of question?
* [ x] **Consistency** - Does the content consistently follow the [style guide](https://o3de.org/docs/contributing/to-docs/style-guide/)?
* [ x] **Help the user** - Does the documentation show the user something *meaningful*?

### DIFF
--- a/content/docs/contributing/to-docs/templates/component-reference-doc.md
+++ b/content/docs/contributing/to-docs/templates/component-reference-doc.md
@@ -258,7 +258,7 @@ Another option is to separate the property tables into [tabs](/docs/contributing
 
 **Examples**:
 
-* [PhysX Collider component](/docs/user-guide/components/reference/physx/collider/): Contains several configurations, which depend on the **Shape** property. Each configuration affect what properties are available, so they are documented in different property tables. 
+* [PhysX Collider component](/docs/user-guide/components/reference/physx/collider/): Contains several configurations, which depend on the **Shape** property. Each configuration affect what properties are available, so they are documented in different property tables separated into tabs. 
 
 * [Camera Rig component](/docs/user-guide/components/reference/camera/camera-rig/): Contains addable property groups with multiple workflows in each. So, each property group is documented in separate property tables and each workflow is further separated into tabs. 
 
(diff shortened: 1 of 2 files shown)


content/docs/contributing/to-docs/templates/component-reference-doc.md
```diff
@@ -257,8 +257,9 @@ For a component card with a complex set of properties, you may need multiple pro
...
 
 **Examples**:
+* [Light component](/docs/user-guide/components/reference/atom/light/): Contains several configurations that don't have a 1-1 relationship with the **Light type** property. Some propery groups are available for more than one light type, so they are documented in different property tables. Each property table specifies which light types support those properties.
 
-* [PhysX Collider component](/docs/user-guide/components/reference/physx/collider/): Contains several configurations, which depend on the **Shape** property. Each configuration affect what properties are available, so they are documented in different property tables. 
+* [PhysX Collider component](/docs/user-guide/components/reference/physx/collider/): Contains several configurations, which depend on the **Shape** property. Each configuration affect what properties are available, so they are documented in different property tables separated into tabs.
```
chanmosq, line 262: Found one of my typos :)

```suggestion
* [PhysX Collider component](/docs/user-guide/components/reference/physx/collider/): Contains several configurations, which depend on the **Shape** property. Each configuration affects what properties are available, so they are documented in different property tables that are separated into tabs. 
```

chanmosq (approved):

FiniteStateGit (approved):

PR #1662, opened 2022-05-23
Update Render Target property description

michabr:
Update the "Render Target" property description of UiImageComponent to reflect the new Atom type.

Signed-off-by: abrmich <abrmich@amazon.com>


willihay:
I don't see this setting change in the current release (22.05). Is this a change that's currently only in the source development branch? If so, this doc change should target the o3de.org development branch instead of main.

michabr:
> I don't see this setting change in the current release (22.05). Is this a change that's currently only in the source development branch? If so, this doc change should target the o3de.org development branch instead of main.

Oh oops. I'll fix that to target development. Thanks.

michabr:
Moved to https://github.com/o3de/o3de.org/pull/1663.

PR #1755, opened 2022-06-24
CMake Essentials Series - Part 1 Blog

aFinchy:
## Change summary

Added Blog CMake Essentials Series - Part 1

### Submission Checklist:

### DIFF
--- a/content/blog/posts/cmake-essentials-series-part-1.md
+++ b/content/blog/posts/cmake-essentials-series-part-1.md
@@ -7,7 +7,7 @@ blog_img: "/images/blog/announcement_thumbnail.jpg"
 full_img: ""
 ---
 
-This series aims to give an overview of some of the most useful functionality in CMake and how to apply it.
+This series aims to give an overview of some of the most useful functionality in CMake and how to apply it in O3DE.
 
 ## Motivation
 
--- a/content/blog/posts/cmake-essentials-series-part-1.md
+++ b/content/blog/posts/cmake-essentials-series-part-1.md
@@ -11,7 +11,7 @@ This series aims to give an overview of some of the most useful functionality in
 
 ## Motivation
 
-CMake is the most widely used build system in the C++ community and understanding how best to use it unlocks a whole host of possibilities when it comes to C++ development. Project setup becomes a breeze and integration with open source projects is made much simpler.
+CMake is the most widely used build system in the C++ community and understanding how best to use it unlocks a whole host of possibilities when it comes to C++ development in O3DE. Project setup becomes a breeze and integration with open source projects is made much simpler.
 
 ## Example
 
--- a/content/blog/posts/cmake-essentials-series-part-1.md
+++ b/content/blog/posts/cmake-essentials-series-part-1.md
@@ -58,7 +58,7 @@ The first line (`#1`) will run the CMake configure step to generate host build f
 
 ## Deliberation
 
-This might seem like mildly more work than opening Visual Studio and creating a project from there, but the massive advantage to this approach is you’ve now created a completely portable C++ program you can build from source on any operating system that supports CMake (most do!). You can easily share this with someone else to quickly build a demo application and even better, with only a few more commands it’s possible to start integrating third-party libraries (no more searching through include paths and linker settings in Visual Studio configuration windows).
+This might seem like mildly more work than opening Visual Studio and creating a project from there, but the massive advantage to this approach is you’ve now created a completely portable C++ program you can build from source on any operating system that supports CMake (most do!). You can easily share this with someone else to quickly build a demo application and even better, with only a few more commands it’s possible to start integrating third-party libraries. (No more searching through include paths and linker settings in Visual Studio configuration windows.)
 
 ## Further Reading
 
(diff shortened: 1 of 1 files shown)


hultonha (commented):
I'm super biased but approving anyway 😅 Thanks so much for getting these published @aFinchy

content/blog/posts/cmake-essentials-series-part-1.md
```diff
@@ -0,0 +1,75 @@
...
+add_executable(${PROJECT_NAME}) # 3
+target_sources(${PROJECT_NAME} PRIVATE main.cpp) # 4
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17) # 5
+```
+
+First we set the CMake version (`#1`). This line always has to come first in a top-level CMakeLists.txt file (`3.15` is a safe bet but feel free to pick a newer version). Next is the `project` command (`#2`), this should immediately follow `cmake_minimum_required` (`LANGUAGES` is optional but good practice). We then create the target to build (`#3`), in this case an executable (`${PROJECT_NAME}` maps to `cmake-essentials-part1`, the name set in the `project` command). Following that we set the sources to be built (`#4`). As with all* `target_` commands me must specify the scope of the items - here we use `PRIVATE` as these files are not going to be relied on by some downstream dependency. Finally we set the C++ version (`#5`), which again is not strictly required but good practice (note we’re not touching compiler flags here, CMake will handle this for us).
+
+*technically not _all,_ but it’s now best practice to always provide this even if legacy commands do work without it.
```
willihay, line 38:
```suggestion
First we set the CMake version (`#1`). This line always has to come first in a top-level CMakeLists.txt file (`3.15` is a safe bet but feel free to pick a newer version).

Next is the `project` command (`#2`), this should immediately follow `cmake_minimum_required` (`LANGUAGES` is optional but good practice).

We then create the target to build (`#3`), in this case an executable (`${PROJECT_NAME}` maps to `cmake-essentials-part1`, the name set in the `project` command).

Following that we set the sources to be built (`#4`). As with all<sup>1</sup> `target_` commands we must specify the scope of the items - here we use `PRIVATE` as these files are not going to be relied on by some downstream dependency.

Finally we set the C++ version (`#5`), which again is not strictly required but good practice (note we’re not touching compiler flags here, CMake will handle this for us).

<sup>1</sup> Technically not _all_ target commands require scope, but it’s best practice to always provide this even if legacy commands work without it.
```

content/blog/posts/cmake-essentials-series-part-1.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+With that setup, the last thing to do is to run CMake.
+
+```bash
+> cmake -S . -B build # 1
+> cmake --build build # 2
+```
+
+The first line (`#1`) will run the CMake configure step to generate host build files (on Windows this will likely default to Visual Studio and can be specified with `-G`). The following line (`#2`) will invoke whatever build system the configure step generated and leave your application in the `build` folder (with Visual Studio this will be `build/Debug/cmake-essentials-part1.exe`).
```
willihay, line 57:
```suggestion
The first line (`#1`) will run the CMake configure step to generate host build files. On Windows this will likely default to Visual Studio and can be specified with `-G`.

The following line (`#2`) will invoke whatever build system the configure step generated and leave your application in the `build` folder. With Visual Studio this will be `build/Debug/cmake-essentials-part1.exe`.
```

content/blog/posts/cmake-essentials-series-part-1.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Deliberation
+
+This might seem like mildly more work than opening Visual Studio and creating a project from there, but the massive advantage to this approach is you’ve now created a completely portable C++ program you can build from source on any operating system that supports CMake (most do!). You can easily share this with someone else to quickly build a demo application and even better, with only a few more commands it’s possible to start integrating third-party libraries (no more searching through include paths and linker settings in Visual Studio configuration windows).
```
willihay, line 61:
```suggestion
This might seem like mildly more work than opening Visual Studio and creating a project from there, but the massive advantage to this approach is you’ve now created a completely portable C++ program you can build from source on any operating system that supports CMake (most do!). You can easily share this with someone else to quickly build a demo application and even better, with only a few more commands it’s possible to start integrating third-party libraries. (No more searching through include paths and linker settings in Visual Studio configuration windows.)
```

willihay (approved):

content/blog/posts/cmake-essentials-series-part-1.md
```diff
@@ -0,0 +1,83 @@
...
+
+The following line (`#2`) will invoke whatever build system the configure step generated and leave your application in the `build` folder. With Visual Studio this will be `build/Debug/cmake-essentials-part1.exe`.
+
+## Deliberation
+
+This might seem like mildly more work than opening Visual Studio and creating a project from there, but the massive advantage to this approach is you’ve now created a completely portable C++ program you can build from source on any operating system that supports CMake (most do!). You can easily share this with someone else to quickly build a demo application and even better, with only a few more commands it’s possible to start integrating third-party libraries. (No more searching through include paths and linker settings in Visual Studio configuration windows.)
+
+## Further Reading
+
+One of the most widely recommended talks on CMake is an excellent presentation by Daniel Pfeifer titled 'Effective CMake'
```
FiniteStateGit, line 73:
```suggestion
One of the most widely recommended talks on CMake is an excellent presentation by Daniel Pfeifer titled 'Effective CMake':
```

FiniteStateGit (changes requested):

FiniteStateGit (approved):

Issue #1686, opened 2022-06-03
[SITE][FEATURE] Left-hand ToC behavior change: Expand section + take user to page


sptramer:
## Describe the feature you'd like available on the website

Currently on the website, the left-hand ToC behavior simply expands a section without loading a new page for the user. Instead, the section should both expand _and_ the user should be taken to the selected page.

## Proposed implementation

In addition to section expansion, any 'click' actions on the left-hand ToC should always bring a user to the selected webpage.

## Available alternatives

No behavior change. This is considered to be an undesirable outcome from a UX standpoint. @o3de/sig-ui-ux has repeatedly given feedback that users expect different behavior.

PR #1859, opened 2022-07-29
Improvements to left docs navigation

chanmosq:
## Change summary

Resolves this issue: https://github.com/o3de/o3de.org/issues/1686

Updated the website to allow the following behaviors in the left-docs-navigation:
- When a user clicks on the section title, they are redirected to that section (`_index.md`) _and_ the section's children pages/sections are displayed. 
- When a user clicks on the "drop-down" icon next to a section, the section's children pages/sections are displayed, _but_ they are not redirected to that section. 

### Submission Checklist:

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -145,8 +145,6 @@ $(function() {
 
 // A helper function to show the list of children in the docs-nav.
 function showChildren(e) {
-  $(e).toggleClass("open");
-
   $(e).siblings('ul').each(function()
   {
     $(this).slideToggle(300);
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -102,15 +102,18 @@ $(function() {
   $("#docs-nav a").each(function(){
     if($(this).attr("href") == url) {
       $(this).addClass("currentPage");
+
+      // Apply attributes to the parent pages/sections for styling.
       $(this).parentsUntil("#docs-nav", "ul").each(function(){
         $(this).addClass("currentAncestor");
         $(this).siblings('a').each(function()
         {
-          $(this).addClass("currentAncestor").addClass("open");
-          
-        });
-        $(this).parent('li').addClass("currentAncestorCategory");
+          $(this).addClass("currentAncestor").addClass("open");          
+        });                  
+        $(this).parent('li').addClass("currentAncestorCategory");    
       })
+      $(this).parent('li').addClass("currentAncestorCategory");    
+      
     }
   })
 });
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -116,7 +116,6 @@ $(function() {
 });
 
 // Applies to the current page in the docs-nav.
-
 $(function() {
   $("#docs-nav a.currentPage").each(function(e){
 
@@ -129,7 +128,8 @@ $(function() {
     }
     $(this).siblings('.has-children-icon').each(function()
     {
-      $(this).addClass("open");      
+      $(this).addClass("open");
+      $(this).addClass("currentPage");
     });
   })
 });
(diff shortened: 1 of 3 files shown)


AMZN-daimini:
This changes has a few minor issues:

### The selection highlight now no longer covers the whole row, as per design
With the new changes: 
image
On o3de.org:
image


### The arrow no longer has a hover state, which makes it harder to understand it's interactable
With the new changes:
hoverState_new
On o3de.org:
hoverState_old

It should be possible to fix both of these issues while still implementing the behavior you seek (separate interaction on title and arrow)

AMZN-daimini:
Additionally, this also seems to be removing the functionality that limits the navigation to the section that is being explored on desktop.
When visiting the User Guide page on o3de.org, only that section is accessible on the left
image

But visiting the same page on the Deploy Preview shows all sections
image

That was a deliberate choice, so it sounds like a regression. My guess is that the changes to the dom structure broke the CSS that hides the other sections.

chanmosq:
Thank you @AMZN-daimini for your feedback!
I've resolved those issues here. 


**Issue**: 
> The selection highlight now no longer covers the whole row, as per design

**Resolved**: 
For the current and parent sections, both their section title and icon now appear highlighted. 
image

**Issue:**
> The arrow no longer has a hover state, which makes it harder to understand it's interactable

**Resolved:**
Icon changes color on hover. It uses the same color as other elements on hover. 


**Issue:** 
> Additionally, this also seems to be removing the functionality that limits the navigation to the section that is being explored on desktop.

**Resolved:**
Only the current section appears.
image

willihay:
I see a few new issues with the latest commits:

image

1) There's extra spacing between many of the topic titles. (I circled the first occurrence in the screenshot.)

2) The section expander arrow's background color/shading is misaligned and produces a different share than the title background. (Circled in the screenshot.)

I think it'd look better if the arrow did not have a background color at all.

willihay (changes requested):
(Requesting visual changes. See screenshot and comments.)

chanmosq:
@willihay Thank you for your look on these updates and for providing feedback!

My latest commit appear to resolve these issues: 
**Issue:**
> 1. As you narrow the width of the browser page, line spacing appears where it's not needed between many of the topic titles. 

**Resolved:** No longer creates space around section titles.
image

**Issue:**
> 2. The section expander arrow's background color/shading is misaligned and produces a different shade than the title background. (Circled in the screenshot.) I think it'd look better if the arrow did not have a background color at all.

**Resolved:** Regarding misalignment, the icon lines up with the text. Same for when the icon is not selected, as shown in the image next to **Project Creation**. When the icon is in an open state, as shown in the image next to **Setup**, the rotation causes the icon to appear lower. So to compensate, the icon is translated up via `translate(-3px, -3px)`. 

The arrow uses the same color as the text, defined by `$o3de-blue` and there's no background color applied to it.

image

chanmosq:
@AMZN-daimini @willihay Can you please check to see if this looks good and can be merged?

willihay:
There's a faint grey background color behind the dropdown arrow (RGB value [240, 242, 244], compared to RGB [248, 251, 254] everywhere else outside of the linkTitle). It's kind of odd-looking because it's a squashed rectangle in my browser (Chrome) - which doesn't even cover the full area behind the dropdown arrow. 

image

Personally, I don't think we want that there, but I'm fine creating an issue and fixing that later.

willihay (approved):

sptramer (approved):
Don't think the RGB/coloration issue is significant enough to request a change.

chanmosq:
Removed the background property! It was applied to both the title link and the drop down link for the page that's currently open. I removed it for both. 

image

willihay (approved):

PR #1884, opened 2022-08-04
Updates to terrain world component docs.

invertednormal:
## Change summary

This PR updates the Terrain World Component documentation to be correct with recent additions and updates the Terrain World Renderer Component documentation to actually have documentation.

The component property descriptions are pulled fairly directly from the edit context tooltips / mins / maxes. Additional context is given in the Usage section.

### Submission Checklist:


content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -11,6 +11,20 @@ The **Terrain World Renderer** component renders the terrain within the bounds s
...
+The terrain renderer component allows terrain to be visualized in the world and controls various global properties that apply to terrain everywhere. 
+
+#### Mesh Configuration
+These properties control the mesh density and blending of terrain LODs (level of detail) and have a very direct impact on vertex performance. The density of the first terrain
+LOD is set by the query resolution in the [Terrain World](/docs/user-guide/components/reference/terrain/world) component, and **First LOD distance** controls how far to render at this
+density. Each subsequent LOD will render for twice the distance at half the resolution all the way until **Mesh render distance** is met. This means that **First LOD distance** actually
+has a much greater impact on total triangle count than the total rendering distance.
+#### Detail Configuration
+These settings impact how far detail materials render, how they fade out, and how they blend with each other. Most of these settings have little impact on performance, however a larger
+render distance here will require more surface data points to be queried as you move around the world and also take more memory.
```
mbalfour-amzn, line 24: Doesn't querying more surface data points directly impact the overall performance?

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -11,6 +11,20 @@ The **Terrain World Renderer** component renders the terrain within the bounds s
...
+#### Clipmap Configuration
+Turning on clipmaps causes all the material blending to be baked into textures for a substantial gain in performance, but also comes with a substantial increase in memory cost.
```
mbalfour-amzn, line 26: From what I've seen so far, I'm not sure it's true that we can say there's a substantial gain in performance. Steady-state rendering performance appears to be faster, but if you fly rapidly around the world, the clipmaps currently seem to have far worse performance than the forward rendering version.

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +36,27 @@ You must add this component to the *Level* entity, the parent of all entities in
 ## Properties
 
 ![Terrain World Renderer component interface.](/images/user-guide/components/reference/terrain/terrain-world-renderer-A.png)
+
+| Property | Description | Values | Default |
+| - | - | - | - |
+| **Mesh Configuration** |
+| Mesh render distance | The distance from the camera that terrain meshes will render. | 1.0 to 100000.0 | 4096.0 |
+| First LOD distance | The distance from the camera that the first Lod renders to. Subsequent LODs will be at double the distance from the previous LOD. | 1.0 to 10000.0 | 128.0 |
```
mbalfour-amzn, line 44: Nit: "Lod" -> "LOD"

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +36,27 @@ You must add this component to the *Level* entity, the parent of all entities in
 ## Properties
 
 ![Terrain World Renderer component interface.](/images/user-guide/components/reference/terrain/terrain-world-renderer-A.png)
+
+| Property | Description | Values | Default |
+| - | - | - | - |
+| **Mesh Configuration** |
+| Mesh render distance | The distance from the camera that terrain meshes will render. | 1.0 to 100000.0 | 4096.0 |
```
mbalfour-amzn, line 43: Suggestion: "will render in meters."

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +36,27 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| First LOD distance | The distance from the camera that the first Lod renders to. Subsequent LODs will be at double the distance from the previous LOD. | 1.0 to 10000.0 | 128.0 |
+| Continuous LOD (CLOD) | Enables the use of continuous level of detail, which smoothly blends geometry between terrain lods. | Boolean | true |
+| CLOD Distance | Distance in meters over which the first lod will blend into the next lod. Subsequent lod blend distances will double with each lod for a consistent visual appearance. | 0.0 to 1000.0 | 16.0
```
mbalfour-amzn, line 46: Nit: "lod" -> "LOD"

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +36,27 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| |
+| **Detail Material Configuration** |
+| Height based texture blending | When turned on, detail materials will use the height texture to aid with blending. | Boolean | 0.0 |
```
mbalfour-amzn, line 49: Default is "False", not "0.0". (Side note - should we change this to default to True?)

content/docs/user-guide/components/reference/terrain/world-renderer.md
mbalfour-amzn: Suggestion: This could also benefit from a brief description of what clipmaps are and how they work. I believe Jiawei posted a helpful diagram on a PR showing what's happening here.

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -11,6 +11,14 @@ The **Terrain World** component provides the data required for other terrain sys
 You must add this component to the *Level* entity, the parent of all entities in an **Open 3D Engine (O3DE)** level.
 {{< /important >}}
 
+## Usage 
+
+The **Min Height** and **Max Height** bounds should be set to as small a range as possible for the most accurate height values. For instance, the terrain renderer uses 16 bit height values, which provides 65536 possible distinct heights. If the difference between the min and max height is 1km, that's a resolution of about 1.5cm.
+
+**Height Query Resolution** is a synchronization point for terrain height used by different terrain systems like physics and rendering. For physics, this controls the density of the underlying height field collider. For rendering, this determines the triangle density of the closest level of detail.
+
+**Surface Data Query Resolution (m)** is similar to above, but it controls the resolution of surface data instead of heights. For instance, this affects the density of detail materials for detail material blending.
```
mbalfour-amzn, line 20: Suggestion: It might be worth describing a bit more how this conceptually creates two terrain world grids - one for height, and one for surface data. All height and surface data providers will be queried at this grid resolution, regardless of the resolution of the data in those providers, so the ideal goal is to use data that matches this resolution. For example, if a heightmap with 10 texels/m is used with a terrain that has a 1 m height query resolution, 9/10 of the texels will be ignored. If the heightmap has 1 texel/m with a 1 m height query resolution, then all of the texels will be used.

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -21,9 +29,10 @@ You must add this component to the *Level* entity, the parent of all entities in
...
-| World Bounds (max) | The maximum value for the worlds bounds. The **max** value must be greater than the **min** value. | Vector3: -2048.0 to 2048.0 | X:`1024.0` Y:`1024.0` Z:`1024.0` |
-| Height Query Resolution (m) | The resolution required between each height sample, in meters. | Vector2: 0.001 to Infinity | X:`1.0` Y:`1.0` |
+| Min Height | The minimum value for the height of terrain. The **min** value must be less than the **max** value.| -65536.0 to 65536.0 | 0.0 |
+| Max Height | The maximum value for the worlds bounds. The **max** value must be greater than the **min** value. | -65536.0 to 65536.0 | 1024.0 |
+| Height Query Resolution (m) | The distance between each height sample position, in meters. | 0.1 to Infinity | 1.0 |
+| Surface Data Query Resolution (m) | The distance between each surface data sample position, in meters. | 0.1 to Infinity | 1.0 |
 
 {{< note >}}
 Because large numbers of terrain height samples can take a long time to calculate, O3DE will not allow the number of samples to exceed 8 million.
```
mbalfour-amzn, line 38: Is this note still accurate?

content/docs/user-guide/components/reference/terrain/world-renderer.md
invertednormal: Which is why I said 

> however a larger render distance here will require more surface data points to be queried as you move around the world and also take more memory.

So the performance is impacted _a little_ as you move the camera, but it's pretty negligible when not moving, and even overall it's not a huge cost to query 1000 surface points every few frames. Doesn't really compare to vertex cost of dense meshes.

content/docs/user-guide/components/reference/terrain/world-renderer.md
invertednormal: I was intentionally light on info with clip maps since I don't know all the details of its implementation. It'd be great for that to be fleshed out by @jiaweig-amzn :). As for the statement about performance, clipmaps _should_ be substantially faster in most normal applications. I'm sure there are many opportunities to make them more performant as time goes on, but maybe we need to be more precise about their expected performance in different applications.

content/docs/user-guide/components/reference/terrain/world-renderer.md
invertednormal: We probably should set this to true, but it's false right now. I'll update the wording.

content/docs/user-guide/components/reference/terrain/world.md
invertednormal: No, I think it'll just crash now if you make the values too crazy. That's what we have an open UX ticket about. I'll remove this.

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -15,9 +15,12 @@ You must add this component to the *Level* entity, the parent of all entities in
 
 The **Min Height** and **Max Height** bounds should be set to as small a range as possible for the most accurate height values. For instance, the terrain renderer uses 16 bit height values, which provides 65536 possible distinct heights. If the difference between the min and max height is 1km, that's a resolution of about 1.5cm.
 
-**Height Query Resolution** is a synchronization point for terrain height used by different terrain systems like physics and rendering. For physics, this controls the density of the underlying height field collider. For rendering, this determines the triangle density of the closest level of detail.
+**Height Query Resolution** and **Surface Data Query Resolution** are a synchronization points for terrain height used by different terrain systems like physics and rendering. The effectively create grids across the world - one for height, and one for surface data. All height and surface data providers will be queried at this grid resolution regardless of the resolution of the data in providers. Ideally, source data in providers with quantized data (like image gradients) will match the query resolution used.
```
mbalfour-amzn, line 18: "are a" -> "are"
"points for terrain height" -> "points for terrain height and surface data"
"The effectively" -> "They effectively"

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -11,6 +11,17 @@ The **Terrain World** component provides the data required for other terrain sys
 You must add this component to the *Level* entity, the parent of all entities in an **Open 3D Engine (O3DE)** level.
 {{< /important >}}
 
+## Usage 
+
+The **Min Height** and **Max Height** bounds should be set to as small a range as possible for the most accurate height values. For instance, the terrain renderer uses 16 bit height values, which provides 65536 possible distinct heights. If the difference between the min and max height is 1km, that's a resolution of about 1.5cm.
```
jiaweig-amzn, line 16: 16 bit -> 16 bits

content/docs/user-guide/components/reference/terrain/world-renderer.md
jiaweig-amzn: I have sketched this out. Feel free to modify any word. The sentence below in brackets is probably too much detail. We can remove that.
Clipmaps are a stack of baked textures centered at the view and covering different distances at different resolutions. Detail materials and macro materials are backed into 2 sets of clipmaps which will take their place when the terrain is rendered. (In O3DE implementation, clipmaps have a uniform size, and the least index stands for the highest resolution clipmap.) Using clipmap will have a substantial gain in performance at the cost of texture memories. Clipmap configuration can be used to find a balance between the performance gain and the memory cost.

jiaweig-amzn (commented):
A general question, we have some CVars that help terrain debugging, including clipmaps', do they go here as well or a different place?

content/docs/user-guide/components/reference/terrain/world.md
invertednormal: In this case "16 bit" is correct

jiaweig-amzn (approved):

content/docs/user-guide/components/reference/terrain/world.md
mbalfour-amzn: Nit: Actually it should be "16-bit". "1km" and "1.5cm" should also have a space between the number and the unit. ("1 km", "1.5 cm")

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -11,6 +11,17 @@ The **Terrain World** component provides the data required for other terrain sys
...
+
+**Height Query Resolution** and **Surface Data Query Resolution** are synchronization points for terrain height and surface data used by different terrain systems like physics and rendering. They effectively create grids across the world - one for height, and one for surface data. All height and surface data providers will be queried at this grid resolution regardless of the resolution of the data in providers. Ideally, source data in providers with quantized data (like image gradients) will match the query resolution used.
+
+For example, if a heightmap with 10 texels/m is used with a terrain that has a 1m height query resolution, 9/10 of the texels will be ignored. If the heightmap has 1 texel/m with a 1 m height query resolution, then all of the texels will be used.
+
+For physics, **Height Query Resolution** controls the density of the underlying height field collider. For rendering, this determines the triangle density of the closest level of detail.
+**Surface Data Query Resolution** affects the density of detail materials for detail material blending.
+
```
mbalfour-amzn, line 24: Suggestion: It would be nice to describe the usage of the CLOD settings here too. Specifically, that the "CLOD distance" should ideally be < 25% of "First LOD distance" and definitely should not be > 50% of "First LOD distance", or else the blend distance will be larger than the distance that the LOD renders for and the blend will never successfully complete, causing visual seams at the LOD boundaries.

mbalfour-amzn (approved):

content/docs/user-guide/components/reference/terrain/world-renderer.md
mbalfour-amzn: At a minimum, I'd suggest rephrasing this as "Most of these settings have little impact on per-frame rendering performance", since there is a noticeable performance impact when moving the camera right now.

chanmosq:
Hey @invertednormal, thanks for updating these docs! 

I've left some suggestions to restructure this doc, so that it matches the format of other current component reference docs. The main change is the removal of the **Usage** section and dispersion of its contents. 

After the changes, it should look like this: 
```
<intro>

## Provider
(As you have it) 

## Depdencies 
(As you have it) 

## Properties
<image of properties>

### Mesh configuration properties
<paragraph, previously in Usage section>

<property table>

### Detail material configuration properties
<paragraph, previously in Usage section>

<property table>

### Clipmap configuration properties
<paragraph, previously in Usage section>

<property table>

```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| - | - | - | - |
+| Mesh render distance | The distance from the camera that terrain meshes will render in meters. | 1.0 to 100000.0 | 4096.0 |
+| First LOD distance | The distance from the camera that the first LOD renders to. Subsequent LODs will be at double the distance from the previous LOD. | 1.0 to 10000.0 | 128.0 |
+| Continuous LOD (CLOD) | Enables the use of continuous level of detail, which smoothly blends geometry between terrain LODs. | Boolean | true |
+| CLOD Distance | Distance in meters over which the first LOD will blend into the next LOD. Subsequent LOD blend distances will double with each LOD for a consistent visual appearance. <br><br>  {{< note >}} Generally this should be less than about 25% of **First LOD distance**. If set too low, seams may appear in the terrain. {{< /note >}} | 0.0 to 1000.0 | 16.0 |
+
+### Detail Configuration
+These settings impact how far detail materials render, how they fade out, and how they blend with each other. Most of these settings have little impact on performance, however a larger
+render distance here will require more surface data points to be queried as you move around the world and also take more memory.
```
chanmosq, line 41:
```suggestion
 These settings impact how far detail materials render, how they fade out, and how they blend with each other. Most of these settings have little impact on performance. However, a larger render distance requires more surface data points to be queried. As you move around the world, this uses more memory.    
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Property | Description | Values | Default |
+| - | - | - | - |
+| Height based texture blending | When turned on, detail materials will use the height texture to aid with blending. | Boolean | False |
+| Detail material render distance (m) | The distance from the camera that the detail material will render in meters. | 1.0 to 2048.0 | 512.0 |
+| Detail material fade distance (m) | The distance over which the detail material will fade out into the macro material. | 0.0 to 2048.0 | 64.0 |
+| Detail material scale | The scale at which all detail materials are rendered at. | 0.0001 to 10000.0 | 1.0 |
+
+### Clipmap Configuration
+
+Clipmaps are a stack of baked textures centered at the view and covering different distances at different resolutions. Detail materials and macro materials are baked into 2 sets of clipmaps which will take their place when the terrain is rendered. Using clipmaps will have a substantial gain in performance at the cost of texture memory. Clipmap configuration can be used to find a balance between the performance gain and the memory cost.
```
chanmosq, line 52: - Wdym by "centered at the view"?
- In my suggested rephrasing, is it accurate to say that the detail and macro materials "replace the clipmaps"?
```suggestion
_Clipmaps_ are a stack of baked textures that are centered at the view. They include baked textures at different distances and at different resolutions. Detail materials and macro materials are baked into two sets of clipmaps. They replace the clipmaps when rendering the terrain. Using clipmaps improves performance, but at the cost of texture memory. You can adjust the clipmap configuration to find a balance between the performance gain and the memory cost. 
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Continuous LOD (CLOD) | Enables the use of continuous level of detail, which smoothly blends geometry between terrain LODs. | Boolean | true |
+| CLOD Distance | Distance in meters over which the first LOD will blend into the next LOD. Subsequent LOD blend distances will double with each LOD for a consistent visual appearance. <br><br>  {{< note >}} Generally this should be less than about 25% of **First LOD distance**. If set too low, seams may appear in the terrain. {{< /note >}} | 0.0 to 1000.0 | 16.0 |
+
+### Detail Configuration
+These settings impact how far detail materials render, how they fade out, and how they blend with each other. Most of these settings have little impact on performance, however a larger
+render distance here will require more surface data points to be queried as you move around the world and also take more memory.
+
+| Property | Description | Values | Default |
+| - | - | - | - |
+| Height based texture blending | When turned on, detail materials will use the height texture to aid with blending. | Boolean | False |
```
chanmosq, line 45:
```suggestion
| **Height based texture blending** | If enabled, detail materials use the height texture for blending. | Boolean | False |
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail material render distance (m) | The distance from the camera that the detail material will render in meters. | 1.0 to 2048.0 | 512.0 |
```
chanmosq, line 46:
```suggestion
| **Detail material render distance** | The distance in meters from the camera that the detail material renders to. | 1.0 to 2048.0 | 512.0 |
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail material fade distance (m) | The distance over which the detail material will fade out into the macro material. | 0.0 to 2048.0 | 64.0 |
```
chanmosq, line 47:
```suggestion
| **Detail material fade distance** | The distance in meters over which the detail material fades out into the macro material. | 0.0 to 2048.0 | 64.0 |
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail material scale | The scale at which all detail materials are rendered at. | 0.0001 to 10000.0 | 1.0 |
```
chanmosq, line 48:
```suggestion
| **Detail material scale** | The scale that all detail materials render at. | 0.0001 to 10000.0 | 1.0 |
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+
+### Clipmap Configuration
+
+Clipmaps are a stack of baked textures centered at the view and covering different distances at different resolutions. Detail materials and macro materials are baked into 2 sets of clipmaps which will take their place when the terrain is rendered. Using clipmaps will have a substantial gain in performance at the cost of texture memory. Clipmap configuration can be used to find a balance between the performance gain and the memory cost.
+
+| Property | Description | Values | Default |
+| - | - | - | - |
+| Clipmap Enabled | When turned on, renders terrain materials with clipmaps instead of rendering the materials directly every frame. | Boolean | false
```
chanmosq, line 56:
```suggestion
| **Clipmap Enabled** | If enabled, renders terrain materials with clipmaps, instead of rendering the materials directly every frame. | Boolean | False
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Clipmap image size | The size of the clipmap image in each layer. | 512, 1024, or 2048 | 1024
```
chanmosq, line 57:
```suggestion
| **Clipmap image size** | The size of the clipmap image in each layer. | 512, 1024, or 2048 | 1024
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Macro clipmap max resolution: texels/m | The resolution of the highest resolution clipmap in the stack. | 0.1 to 100.0 | 2.0
```
chanmosq, line 58: Is it possible to remove "texels/m" in the property name of the component card?
```suggestion
| **Macro clipmap max resolution: texels/m** | The resolution of the highest resolution clipmap in the stack. | 0.1 to 100.0 | 2.0
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail clipmap max resolution: texels/m | The resolution of the highest resolution clipmap in the stack. | 10.0 to 4096.0 | 2048.0
```
chanmosq, line 59:
```suggestion
| **Detail clipmap max resolution: texels/m** | The resolution of the highest resolution clipmap in the stack. | 10.0 to 4096.0 | 2048.0
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Macro clipmap scale base | The scale base between two adjacent clipmap layers. For example, 3 means the (n+1)th clipmap covers 3^2 = 9 times the area covered by the nth clipmap. | 1.1 to 10.0 | 2.0
```
chanmosq, line 60:
```suggestion
| **Macro clipmap scale base** | The scale base between two adjacent clipmap layers of macro materials. For example, 3 means that the (n+1)<sup>th</sup> clipmap covers 3<sup>2</sup> = 9 times the area that's covered by the n<sup>th</sup> clipmap. | 1.1 to 10.0 | 2.0
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail clipmap scale base | The scale base between two adjacent clipmap layers. For example, 3 means the (n+1)th clipmap covers 3^2 = 9 times the area covered by the nth clipmap. | 1.1 to 10.0 | 2.0
```
chanmosq, line 61:
```suggestion
| **Detail clipmap scale base** | The scale base between two adjacent clipmap layers of detail materials. For example, 3 means that the (n+1)<sup>th</sup> clipmap covers 3<sup>2</sup> = 9 times the area that's covered by the n<sup>th</sup> clipmap. | 1.1 to 10.0 | 2.0
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Macro clipmap margin size: texels | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates but also results in lower resolution clipmaps rendering closer to the camera. | 1 to 16 | 4
```
chanmosq, line 62:
```suggestion
| **Macro clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 to 16 | 4
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,42 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Detail clipmap margin size: texels | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates but also results in lower resolution clipmaps rendering closer to the camera. | 1 to 16 | 4
```
chanmosq, line 63:
```suggestion
| **Detail clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 to 16 | 4
```

chanmosq (changes requested):
Made some suggestions for `world-render.md`

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -21,10 +21,19 @@ You must add this component to the *Level* entity, the parent of all entities in
...
 
-{{< note >}}
-Because large numbers of terrain height samples can take a long time to calculate, O3DE will not allow the number of samples to exceed 8 million.
-{{< /note >}}
+
+## Usage 
+
+The **Min Height** and **Max Height** bounds should be set to as small a range as possible for the most accurate height values. For instance, the terrain renderer uses 16-bit height values, which provides 65536 possible distinct heights. If the difference between the min and max height is 1 km, that's a resolution of about 1.5 cm.
+
+**Height Query Resolution** and **Surface Data Query Resolution** are synchronization points for terrain height and surface data used by different terrain systems like physics and rendering. They effectively create grids across the world - one for height, and one for surface data. All height and surface data providers will be queried at this grid resolution regardless of the resolution of the data in providers. Ideally, source data in providers with quantized data (like image gradients) will match the query resolution used.
```
chanmosq, line 34: What is meant by "synchronization points" and "providers"? "Providers" may be ambiguous. 

Similarly, "source data" may be ambiguous, so I updated the last sentence. Can you verify if that's right?

```suggestion
 **Height Query Resolution** and **Surface Data Query Resolution** are synchronization points for both terrain height and surface data. Different terrain systems, like physics and rendering, use them to create height and surface data grids across the world. All height and surface data providers are queried at this grid resolution, regardless of the resolution of the data in providers.  
 
We recommend that heightmaps or surface data that have quantized source data, like image gradients, match the Height Query Resolution.
```

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -21,10 +21,19 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+
+For example, if a heightmap with 10 texels/m is used with a terrain that has a 1m height query resolution, 9/10 of the texels will be ignored. If the heightmap has 1 texel/m with a 1 m height query resolution, then all of the texels will be used.
```
chanmosq, line 36:
```suggestion
For example, if a terrain uses a heightmap with 10 texels/meter with a 1 meter height query resolution, then 9 out of 10 of the texels are ignored. If the heightmap has 1 texel/meter with a 1 meter height query resolution, then all of the texels are used.
```

chanmosq (changes requested):

content/docs/user-guide/components/reference/terrain/world-renderer.md
invertednormal: The way the last two sentences read changes the meaning a bit. More surface points are queried when you move around the world, and it takes more memory (regardless of if you're moving or not). Changing to 
"However, a larger render distance requires more surface data points to be queried as you move around the world, and it also uses more memory."

content/docs/user-guide/components/reference/terrain/world-renderer.md
invertednormal: This needed a bit more work, I messaged you directly with a suggestion.

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -21,10 +21,19 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Height Query Resolution (m) | The distance between each height sample position, in meters. | 0.1 to Infinity | 1.0 |
+| Surface Data Query Resolution (m) | The distance between each surface data sample position, in meters. | 0.1 to Infinity | 1.0 |
 
-{{< note >}}
-Because large numbers of terrain height samples can take a long time to calculate, O3DE will not allow the number of samples to exceed 8 million.
-{{< /note >}}
+
+## Usage 
+
+The **Min Height** and **Max Height** bounds should be set to as small a range as possible for the most accurate height values. For instance, the terrain renderer uses 16-bit height values, which provides 65536 possible distinct heights. If the difference between the min and max height is 1 km, that's a resolution of about 1.5 cm.
```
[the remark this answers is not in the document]
invertednormal: Your wording improves things :)

To try to explain it more, imagine a terrain with a min height of 0 and max height of 1000. With 65536 total possible distinct values to work with, that means with a 1000-meter range there are enough positions to have one distinct position every 1.5 cm. (The actual math is just 1000m / 65535 = 0.015259m). So, we could represent 0 perfectly, but 1cm would get rounded up to 1.5cm and 0.5cm would round down to 0cm. If your total height range was higher, like 10,000m (say you wanted to handle sea level to mount Everest) then you'd only have one distinct value per 15cm, so there'd be more rounding, and you might lose some fine detail.

content/docs/user-guide/components/reference/terrain/world.md
invertednormal: "Synchronization points" - The query resolutions tell other systems that rely on terrain how to query data for their system. I referred to them as synchronization points because it helps keep different terrain systems in sync. As an example, if the query resolution is 1m, then both rendering and physics will query the same points 1 meter apart for their data, and everything will like up - a ball will bounce off the terrain in the location that it looks like it should bounce. Without this property, we'd have to leave it up to the user to make sure they made everything sync up by themselves. So, they don't create actual grids across the world, it's just a way of understanding how they work (ie query in regular intervals in x and y according to the query resolution)

"Providers" - this just means any component which can respond to queries about heights or surface data - it "provides" data for the terrain system.

"Source data" I think your clarification works. Source data could provide to anything providing the height or surface information like an image or noise function.

invertednormal: How about this:

> **Height Query Resolution** and **Surface Data Query Resolution** are synchronization points for both terrain height and surface data. All of the entities providing input height and surface data for the terrain system will be queried at this grid resolution regardless of the resolution of the data contained in those entities. We recommend that heightmaps or surface data that have quantized source data, like image gradients, match the Height Query Resolution.

content/docs/user-guide/components/reference/terrain/world.md
invertednormal: After reading this again I'm going to change "heightmap" to "heightmap image" to be clearer since we're talking about an image source that happens to contain heightmap information.

content/docs/user-guide/components/reference/terrain/world.md
invertednormal: This actually provoked a long discussion with @mbalfour-amzn and this is what we landed on:

> **Height Query Resolution** and **Surface Data Query Resolution** are used to define consistent spacing, centered at the origin, for terrain height and surface queries across the entire world. Different systems that use terrain data, such as physics and rendering, can also use these conceptual grids to make implementation decisions about how to best use the terrain data. The input data to the terrain system can be authored or generated at any resolution, but the terrain system will by default query the input data only at points aligned to the grid. This creates a consistent view of the input data. Since the input data will normally only be used at these resolutions, we recommend that input gradients that have quantized source data, like images, match the appropriate query resolution.

chanmosq: Great! This makes a lot more sense. And "conceptual grid" is a good way to describe it.

```suggestion
**Height Query Resolution** and **Surface Data Query Resolution** are used to define consistent spacing, centered at the origin, for terrain height and surface queries across the entire world. Different systems that use terrain data, such as physics and rendering, can also use these conceptual grids to make implementation decisions about how to best use the terrain data. By default, the terrain system queries the input data only at points aligned to the grid. However, you can author or generate input data to the terrain system at any resolution. This creates a consistent view of the input data. Since the input data is only used at these resolutions, we recommend that input gradients that have quantized source data, like images, match the appropriate query resolution.
```

content/docs/user-guide/components/reference/terrain/world.md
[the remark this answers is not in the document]
chanmosq: Cool! I incorporated some of your explanation into my suggestion: 
```suggestion
We recommend that you set the **Min Height** and **Max Height** bounds to as small of a range as possible,  to achieve more accurate height values . The terrain renderer uses 16-bit height values, which provides 65536 possible distinct heights. To get the resolution, calculate: (Max Height - Min Height) / 65536. So, if the height range is 1 km, then the resolution is about 1.5 cm. This results in finer height detail. 
```

content/docs/user-guide/components/reference/terrain/world.md
```diff
@@ -21,10 +21,19 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| Height Query Resolution (m) | The distance between each height sample position, in meters. | 0.1 to Infinity | 1.0 |
+| Surface Data Query Resolution (m) | The distance between each surface data sample position, in meters. | 0.1 to Infinity | 1.0 |
 
-{{< note >}}
-Because large numbers of terrain height samples can take a long time to calculate, O3DE will not allow the number of samples to exceed 8 million.
-{{< /note >}}
+
+## Usage 
+
+We recommend that you set the **Min Height** and **Max Height** bounds to as small of a range as possible,  to achieve more accurate height values . The terrain renderer uses 16-bit height values, which provides 65536 possible distinct heights. So, if the difference between the min and max height is 1 km, then the resolution is about 1.5 cm.
```
chanmosq, line 32: I incorporated your explanations into this suggestion: 

```suggestion
We recommend that you set the **Min Height** and **Max Height** bounds to as small of a range as possible,  to achieve more accurate height values . The terrain renderer uses 16-bit height values, which provides 65536 possible distinct heights. To get the resolution, calculate: (Max Height - Min Height) / 65536. So, if the height range is 1 km, then the resolution is about 1.5 cm. This results in finer height detail. 
```

content/docs/user-guide/components/reference/terrain/world.md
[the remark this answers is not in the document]
chanmosq: Cool! I incorporated this into my [latest suggestion](https://github.com/o3de/o3de.org/pull/1884#discussion_r942946590).

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,45 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| **Clipmap image size** | The size of the clipmap image in each layer. | 512, 1024, or 2048 | 1024
+| **Macro clipmap max resolution: texels/m** | The resolution of the highest resolution clipmap in the stack. | 0.1 to 100.0 | 2.0
+| **Detail clipmap max resolution: texels/m** | The resolution of the highest resolution clipmap in the stack. | 10.0 to 4096.0 | 2048.0
+| **Macro clipmap scale base** | The scale base between two adjacent clipmap layers of macro materials. For example, 3 means that the (n+1)<sup>th</sup> clipmap covers 3<sup>2</sup> = 9 times the area that's covered by the n<sup>th</sup> clipmap. | 1.1 to 10.0 | 2.0
+| **Detail clipmap scale base** | The scale base between two adjacent clipmap layers of detail materials. For example, 3 means that the (n+1)<sup>th</sup> clipmap covers 3<sup>2</sup> = 9 times the area that's covered by the n<sup>th</sup> clipmap. | 1.1 to 10.0 | 2.0
+| **Macro clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 to 16 | 4
```
FiniteStateGit, line 54:
```suggestion
| **Macro clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 - 16 | `4` |
```

content/docs/user-guide/components/reference/terrain/world-renderer.md
```diff
@@ -22,3 +22,45 @@ You must add this component to the *Level* entity, the parent of all entities in
...
+| **Detail clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 to 16 | 4
```
FiniteStateGit, line 55:
```suggestion
| **Detail clipmap margin size: texels** | The margin of the clipmap beyond the visible data. Increasing the margins results in less frequent clipmap updates, but also results in lower resolution clipmaps when rendering closer to the camera. | 1 - 16 | `4` |
```

FiniteStateGit (changes requested):

chanmosq (approved):

FiniteStateGit (approved):

PR #1433, opened 2022-02-21
Added Chromatic Aberration Doc

tjmgd:
Signed-off-by: T.J. McGrath-Daly <tj.mcgrath.daly@huawei.com>

### Submission Checklist:

* [ ] **Descriptive active voice** - Do descriptive sentences have a clear *subject* and *action verb*?
* [ ] **Answer the question at hand** - Does the documentation answer a *what*, *why*, *how*, or *where* type of question?
* [ ] **Consistency** - Does the content consistently follow the [style guide](https://o3de.org/docs/contributing/to-docs/style-guide/)?
* [ ] **Help the user** - Does the documentation show the user something *meaningful*?

PR #1949, opened 2022-08-31
add chromatic aberration component ref

FiniteStateGit:
Add basic docs and example image for the chromatic aberration component based on contributor PR https://github.com/o3de/o3de.org/pull/1433.


willihay (approved):

micronAMZN (changes requested):
This needs to be added to the component reference index as well.

docs/user-guide/components/reference/

micronAMZN (approved):

PR #1970, opened 2022-09-13
O3DE user guide updated with new physics materials workflow.

moraaar:
Signed-off-by: moraaar <moraaar@amazon.com>

## What’s changing?
For next O3DE release the Physics material assets will be managed independently. The concept of physics material library has been removed and now one asset is one physics material. A physics material is created using Asset Editor tool.

## Why is this change being made?
To simplify the workflow of authoring and using physics materials, we're making it more like how render materials work. Having a library asset added an extra layer of complexity that users had to deal with and learn how to use within the editor. Now physics materials can be assigned to collider components directly like any other asset.

### Submission Checklist:


content/docs/user-guide/interactivity/physics/nvidia-physx/_index.md
```diff
@@ -11,10 +11,9 @@ O3DE's PhysX system acts upon entities to create realistic physical effects such
 + [PhysX Gems](#physx-gems)
 + [PhysX Components](#physx-components)
 + [PhysX Configuration](#physx-configuration)
-+ [PhysX Materials](#physx-materials)
++ [Physics Materials](#physics-materials)
```
hultonha, line 14: question: This might be fine but as there's only 1 Physics here vs the other PhysX ... it looks a bit odd (even if technically correct). I just wanted to mention maybe we should move these to a separate section potentially? Or just move from being in the middle? Not 100% sure on the best fix but I think it could lead to confusion as-is potentially

hultonha (approved):
Updates look excellent Aaron, screen shots are really good too. I only have one minor bit of feedback but otherwise looks great! 🥳

content/docs/user-guide/interactivity/physics/nvidia-physx/_index.md
moraaar: Yeah, I spotted that too, I thought about changing it to PhysX Material, but everywhere in the documentation is written as "physics material" and I didn't want to break that consistency, using both terms will leading to more confusion and make user searches more difficult, so I decided to leave it as it is.

I could move the point to the middle, but relevance wise it's in the right spot in the list, so I rather keep it even though is not as pretty to the eye.

If anybody feels strongly that we should change it, please let me know.

hultonha: Yeah I totally understand, keeping it called 'Physics' definitely makes sense, it's just the placement that looks out of place. It's not a huge deal though so I think keeping as-is for now is okay (could maybe ask docs about restructuring a little as a separate PR to maybe have a PhysX and Physics section in future?).

micronAMZN: If the proper name is Physics Materials, we should go with that.

micronAMZN (approved):

PR #1959, opened 2022-09-07
HttpRequestorGem: Add instructions about how to setup runtime dependencies on gem

lmbr-pip:
## Change summary

Was asked for by contributor, fixes https://github.com/o3de/o3de.org/issues/1920

* small tweaks to http requestor documentation

### Submission Checklist:

* [x] **Descriptive active voice** - Do descriptive sentences have a clear *subject* and *action verb*?
* [x] **Answer the question at hand** - Does the documentation answer a *what*, *why*, *how*, or *where* type of question?
* [x] **Consistency** - Does the content consistently follow the [style guide](https://o3de.org/docs/contributing/to-docs/style-guide/)?
* [x] **Help the user** - Does the documentation show the user something *meaningful*?

Issue #1920, opened 2022-08-18
[ISSUE][DOCS] HttpRequestor - Expand documentation to cover component ordering


lmbr-pip:
## Describe the issue

Dev was confused about how to use HttpRequestor due to lack of awareness of component ordering: https://github.com/o3de/o3de/issues/8351.

Add to the HttpRequestor docs:
* How to declare a dependency on the component so users can understand when its activated.
    * Note about when HttpRequestor is activated generally in the startup sequence
* Note about recent change to control behaviour on EC2 instances
* Add component diagram as per https://docs.unity3d.com/Manual/ExecutionOrder.html%5B%5D(https://unity3d.com/files/images/ogimg.jpg)

## Impacted page(s)

https://www.o3de.org/docs/user-guide/gems/reference/network/http-requestor/ 

## Expected experience
Page resolves confusion expressed by user trying to use gem.

FiniteStateGit:
@lmbr-pip Are you planning to work on these documentation changes or would you be providing more information on the changes you are requesting?

lmbr-pip:
Fixed by https://github.com/o3de/o3de.org/pull/1989

PR #1989, opened 2022-09-22
HttpRequestor: Add instructions about creating dependency on gem and deactivting EC2 IMDS

lmbr-pip:
Updated version of https://github.com/o3de/o3de.org/pull/1959

* Add documentation around dependency setup, asked for in https://github.com/o3de/o3de.org/issues/1920
     * Retargets for stabilization this PR that originally had changes https://github.com/o3de/o3de.org/pull/1959
* Adds note about preventing calls to EC2 metadata service, as reported in #sig-network chat.
     * Repeats information in https://github.com/o3de/o3de.org/pull/1991 so would be good to standardized language between the two changes (in this case probably need to repeat information as HttpRequestor is used by non AWS users, who will be surprised by the need to set this variable)

Signed-off-by: Pip Potter <61438964+lmbr-pip@users.noreply.github.com>

### Submission Checklist:

### DIFF
--- a/content/docs/user-guide/gems/reference/network/http-requestor.md
+++ b/content/docs/user-guide/gems/reference/network/http-requestor.md
@@ -39,7 +39,7 @@ void AutomatedTestingSystemComponent::GetRequiredServices(AZ::ComponentDescripto
 
 ### Turn off AWS EC2 Instance Metadata Service calls
 
-The HttpRequestor gem uses the [AWS C++ SDK](https://github.com/aws/aws-sdk-cpp) under-the-hood to provide the Http(s) client. The code is configured to not require any AWS credentials or account information.
+The HttpRequestor Gem uses the [AWS C++ SDK](https://github.com/aws/aws-sdk-cpp) to provide the Http(s) client. You don't need to provide AWS credentials or account information to use this Gem.
 
 However, if you are not running on AWS EC2 compute its recommended that you turn off the [AWS_EC2_METADATA_DISABLED](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L104) environment variable. This will prevent any reach out the AWS EC2 Instance Metadate Service (IMDS), which may occur to retrieve configuration, region and credential information. Requests to EC2 IMDS will fail on non EC2 compute leading to delays and wasted network resources.
 
--- a/content/docs/user-guide/gems/reference/network/http-requestor.md
+++ b/content/docs/user-guide/gems/reference/network/http-requestor.md
@@ -11,7 +11,7 @@ The HTTPRequestor Gem provides functionality to make asynchronous HTTP/HTTPS req
 This feature is currently supported only on Windows, it could be expanded to other platforms.
 {{< /note >}}
 
-## Getting Started
+## Getting started
 
 To use the HttpRequestor Gem, it must be enabled in the project. For more information refer to [Adding and Removing Gems in a Project](/docs/user-guide/project-config/add-remove-gems/).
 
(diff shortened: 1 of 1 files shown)


junbo75 (approved):

content/docs/user-guide/gems/reference/network/http-requestor.md
```diff
@@ -25,6 +25,29 @@ BUILD_DEPENDENCIES
...
+   ...
+   required.push_back(AZ_CRC_CE("HttpRequestorService"));
+}
+```
+
+### Turn off AWS EC2 Instance Metadata Service calls
+
+The HttpRequestor gem uses the [AWS C++ SDK](https://github.com/aws/aws-sdk-cpp) under-the-hood to provide the Http(s) client. The code is configured to not require any AWS credentials or account information.
+
+However, if you are not running on AWS EC2 compute its recommended that you turn off the [AWS_EC2_METADATA_DISABLED](https://github.com/aws/aws-sdk-cpp/blob/main/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L104) environment variable. This will prevent any reach out the AWS EC2 Instance Metadate Service (IMDS), which may occur to retrieve configuration, region and credential information. Requests to EC2 IMDS will fail on non EC2 compute leading to delays and wasted network resources.
```
willihay, line 44: AWS branding uses "Amazon EC2" instead of "AWS EC2", so I recommend sticking to brand here.
Also, I'd recommend linking to the description of the env variable instead of the source code.
```suggestion
However, if your project is not running on Amazon EC2 compute, it's recommended that you turn off the [AWS_EC2_METADATA_DISABLED](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) environment variable. This will prevent any calls to the Amazon EC2 Instance Metadata Service (IMDS) from attempting to retrieve configuration, region, and credential information. Requests to IMDS will fail if you are not using EC2, leading to delays and wasted network resources.
```

willihay (changes requested):

content/docs/user-guide/gems/reference/network/http-requestor.md
```diff
@@ -25,6 +25,29 @@ BUILD_DEPENDENCIES
...
+
+```cpp
+void AutomatedTestingSystemComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
+{
+   ...
+   required.push_back(AZ_CRC_CE("HttpRequestorService"));
+}
+```
+
+### Turn off AWS EC2 Instance Metadata Service calls
```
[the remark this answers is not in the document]
lmbr-pip: Good catch, always forget EC2 is Amazon EC2 branded

content/docs/user-guide/gems/reference/network/http-requestor.md
lmbr-pip: Technically that link is to the AWS CLI, which I didn't use as I thought it was confusing because we are not configuring the AWS CLI but the AWS C++ SDK. Unfortunately the AWS C++ SDK docs do not cover this env var doc.

willihay (approved):

content/docs/user-guide/gems/reference/network/http-requestor.md
```diff
@@ -25,6 +25,29 @@ BUILD_DEPENDENCIES
...
+   ...
+   required.push_back(AZ_CRC_CE("HttpRequestorService"));
+}
+```
+
+### Turn off Amazon EC2 Instance Metadata Service calls
+
+The HttpRequestor Gem uses the [AWS C++ SDK](https://github.com/aws/aws-sdk-cpp) to provide the Http(s) client. You don't need to provide AWS credentials or account information to use this Gem.
+
+However, if your project is not running on Amazon EC2 compute, it's recommended that you turn off the [AWS_EC2_METADATA_DISABLED](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) environment variable. This will prevent any calls to the Amazon EC2 Instance Metadata Service (IMDS) from attempting to retrieve configuration, region, and credential information. Requests to IMDS will fail if you are not using EC2, leading to delays and wasted network resources.
```
allisaurus, line 44: I know this comment is late and a bit of a nitpick but it should probably be phrased as "turn _on_ the AWS_EC2_METADATA_DISABLED environmental variable..." since it gets set to `true`.

PR #1997, opened 2022-09-27
Make Visual Studio examples more version agnostic where possible

willihay:
Signed-off-by: Willow Hayward <17654918+willihay@users.noreply.github.com>

## Change summary

In general, build directory examples for Visual Studio builds should not contain a version reference (example: `build\windows_vs2019`). Project Manager does not use it, and having build examples that do can lead to confusion and duplicate build directories. Instead, use just the platform, such as `build\windows`.

Also, in some places we can say simply "Visual Studio" instead of "Visual Studio 2019". This is important as we support Visual Studio 2022.

Adds a table of common cmake generators to the **Configure and Build** topic.

In CLI examples that use the "Visual Studio 16" generator, add a note that links to the table of common generators.

### Submission Checklist:


micronAMZN (approved):
👍

content/docs/user-guide/build/configure-and-build.md
```diff
@@ -64,7 +71,7 @@ The typical CMake command used to configure a project looks like the following:
 {{% tab name="Windows" %}}
 
 ```cmd
-cmake -B build/windows_vs2019 -S . -G "Visual Studio 16" -DLY_3RDPARTY_PATH=<downloadable-packages-directory>
+cmake -B build/windows -S . -G "Visual Studio 16" -DLY_3RDPARTY_PATH=<downloadable-packages-directory>
```
sptramer, line 74: Minor nit that we can handle later if needed - should we have three tabs here, two for Windows with different VS versions? Or should we just update to "Visual Studio 17" in stabilization / at the time of merge?

sptramer (approved):

content/docs/user-guide/build/configure-and-build.md
willihay: Having 3 tabs is not a bad idea. I'd prefer that. I'll create a GHI for follow up.

willihay: https://github.com/o3de/o3de.org/issues/2026

PR #2007, opened 2022-09-30
Version info updates for 22.10

willihay:
## Change summary

Updates to version requirements for:

- Visual Studio
- CMake
- Linux (including CMake and Clang info)
- macOS/iOS (where to find requirements and support info)

### Submission Checklist:

@netlify /docs/welcome-guide/requirements


willihay:
Addresses https://github.com/o3de/o3de/pull/7714

micronAMZN (approved):

spham-amzn:
We still supported Ubuntu 20.04, but with 22.10 release we want to indicate that we support 22.04 as well, but right now that version is considered experimental (much like or MacOS support), which is reasonable since 22.04 was only released middle of this year. I'm not sure of a good approach or location on how to message that. 

Related to that, the instructions for getting the latest version of cmake only Ubuntu is 20.04 specific, so I think that section can be simplified by just redirecting them to the official CMake download page which will provide instructions on how to install the latest version for any of their supported operation systems (https://cmake.org/download/)

spham-amzn:
The instructions for installing clang are also ubuntu 20.04 specific. On ubuntu 22.04, all you need is ```sudo apt-get install clang``` (without the version number) because by default this gives you clang-14 which is supported

FiniteStateGit (approved):

content/docs/welcome-guide/requirements.md
```diff
@@ -145,17 +171,43 @@ cmake --version
...
+
+```shell
+sudo apt-get install clang
+```
+
+After Clang is installed, it needs to be added to `/etc/environment` as two new environment variables. For example, if you installed clang-14, add the following:
+
+```shell
+CC="/usr/bin/clang-14"
+CXX="/usr/bin/clang++-14"
```
spham-amzn, line 205: This actually is no longer needed, even for clang-12

content/docs/welcome-guide/requirements.md
```diff
@@ -145,17 +171,43 @@ cmake --version
...
 sudo apt-get install clang-12 
 ```
 
-After clang-12 is installed, it needs to be added to `/etc/environment` as two new environment variables.
+After Clang is installed, it needs to be added to `/etc/environment` as two new environment variables. For example, if you installed clang-12, add the following:
 
 ```shell
 CC="/usr/bin/clang-12"
 CXX="/usr/bin/clang++-12"
 ```
```
spham-amzn, line 190: This is no longer needed

spham-amzn, line 190: the CC, CXX environment setup is no longer needed

content/docs/welcome-guide/requirements.md
```diff
@@ -145,17 +171,43 @@ cmake --version
...
+```shell
+sudo apt-get install clang
+```
+
+After Clang is installed, it needs to be added to `/etc/environment` as two new environment variables. For example, if you installed clang-14, add the following:
+
+```shell
+CC="/usr/bin/clang-14"
+CXX="/usr/bin/clang++-14"
+```
```
spham-amzn, line 206: These instructions are no longer needed

content/docs/welcome-guide/requirements.md
```diff
@@ -113,7 +116,13 @@ If the current CMake version was not returned because CMake cannot be found, loc
 
 ## Linux
 
-The primary Linux distribution for using the O3DE Editor is Ubuntu {{< versions/ubuntu >}}. The following instructions describe how to retrieve and install the required software packages through Ubuntu's `apt-get` command-line utility.
+The primary Linux distribution for using the O3DE Editor is Ubuntu {{< versions/ubuntu >}}.
+
+{{< note >}}
+Support for Ubuntu 22.04 in O3DE is considered experimental.
```
spham-amzn, line 122: minor: Maybe this can be worded the same way we did for Mac:
```suggestion
Support for Ubuntu 22.04 is in an experimental stage.
```

spham-amzn (approved):

spham-amzn (approved):

PR #2009, opened 2022-09-30
Update API reference for 22.10.

sptramer:
API reference update for the 22.10 release. Please spot-check for:

* No broken links
* Version numbers are correct
* Landing page copy looks correct


micronAMZN (approved):

willihay:
> No broken links

Do we have some automation for that now? I wonder if the tool @FiniteStateGit was just using could help here.

FiniteStateGit:
@willihay I'll run it against the preview branch today.

static/docs/api/frameworks/atomcore/annotated.html
```diff
@@ -20,7 +20,7 @@
   <td id="projectlogo"><img alt="Logo" src="/images/api-reference/api-ref-icon.png"/></td>
   <td id="projectalign" style="padding-left: 0.5em;">
    <div id="projectname">Open 3D Engine AtomCore API Reference
-   &#160;<span id="projectnumber">2111.2</span>
+   &#160;<span id="projectnumber">2210.0</span>
```
willihay, line 23: Now that we've switched to a XX.XX scheme for the friendly name, this should probably be 22.10.0.

static/docs/api/frameworks/atomcore/class_a_z_1_1_data_1_1_instance_data-members.html
```diff
@@ -20,7 +20,7 @@
   <td id="projectlogo"><img alt="Logo" src="/images/api-reference/api-ref-icon.png"/></td>
```
willihay, line 20: In the future, it'd be nice if this logo had a clickable link to take you back to the [API Reference homepage](https://www.o3de.org/docs/api/frameworks/). As it is, there's no way to navigate to any other API other than using the browser back button. I realize that the API Reference is already scheduled for an overhaul. I just mention it in case it's a quick interim fix. Filed as a GHI: https://github.com/o3de/o3de.org/issues/2012

willihay (changes requested):

FiniteStateGit:
[PR2009-broken-links.csv](https://github.com/o3de/o3de.org/files/9688005/PR2009-broken-links.csv)

I can investigate these broken links when I review the PR.

sptramer:
> [PR2009-broken-links.csv](https://github.com/o3de/o3de.org/files/9688005/PR2009-broken-links.csv)
> 
> I can investigate these broken links when I review the PR.

Now that's interesting! Let me run my local broken-link-checker and see what comes out of it. There's some real possibilities that a few classes got removed in the latest release and this could be leading to broken links.

sptramer:
My local broken link check with [mlc](https://github.com/becheran/mlc) showed only the following page missing: `class_az_framework_1_1_spawnable_1_1_entity_alias_visitor_base.html`. It could be the case that this is a private class of some kind; Will perform further investigation, because this could be a breakage issue in stabilization.

sptramer:
OK, this is a strange one - I can't get the missing page generated without going down to the `private` level of the Framework, which makes sense. _However_, we definitely don't want all private APIs in public, that's way too much to offer for people who would mostly be interested in treating the Frameworks and Gems as SDKs. We can ship _without_ this reference generated for 22.10 if the SIG is okay with that, and then do a post-release investigation into why the error occurs when we get to reference generation.

willihay:
> We can ship _without_ this reference generated for 22.10 if the SIG is okay with that, and then do a post-release investigation into why the error occurs when we get to reference generation.

I'm okay with that.

sptramer:
Created https://github.com/o3de/o3de.org/issues/2022 for addressing the problem where we have public classes inheriting from protected/private which give a lower-level abstraction but may still be exposing public functionality that's considered important.

willihay (approved):

sptramer:
@willihay @micronAMZN +2 ✔️ , go ahead and push the green button! We really need to set up auto-merge after access control is in place.

PR #2078, opened 2022-11-10
Add a set of recommended hardware requirements

FiniteStateGit:
@netlify /docs/welcome-guide/requirements/

## Change summary

- Changes to minimum requirements:
    - Increase RAM to 16GB (with note that 8GB can be manageable by limiting CPU processes).
    - Remove "or better" verbiage because recommended specs are now available.
    - CPU should be 64-bit x86-capable and support the SSE 4.1 SIMD instruction set.
    - Note difference in disk space requirements for Installer vs. source build.
    - Note requirement of 2GB per core in addition to existing system memory needs required to compile from source.
    - Remove screen resolution requirements.
    - Remove Metal-compatible from GPU reqs.

- Add recommended specs:

| Hardware Component | Recommended Specification |
| - | - |
| CPU | Hexa-core (6 cores) 64-bit x86-capable Intel or AMD processor, 2.5 GHz, supports the SSE 4.1 SIMD instruction set |
| RAM | 32 GB RAM |
| GPU | - DirectX 12 or Vulkan-compatible video card <br> - 6 GB VRAM <br> - Shader Model 6.2 (or Shader Model 6.3 to use raytracing features) <br> - NVIDIA GeForce GTX 16 Series, driver version 471.11 or later recommended _or_ <br> - AMD RX 5000 Series |
| Hard Disk | SSD with 1 TB of free disk space |

### Submission Checklist:

### DIFF
--- a/content/docs/welcome-guide/requirements.md
+++ b/content/docs/welcome-guide/requirements.md
@@ -6,29 +6,40 @@ weight: 300
 toc: true
 ---
 
-**Open 3D Engine (O3DE)** has a minimum set of hardware and software requirements for development, as outlined in the following sections. In the [Software
-prerequisites and configuration](#software-prerequisites)
-section, each prerequisite is listed along with any specific configuration steps
-required.
+**Open 3D Engine (O3DE)** has a minimum set of hardware and software requirements for development, as outlined in the following sections. In the [Software prerequisites and configuration](#software-prerequisites) section, each prerequisite is listed along with any specific configuration steps required.
 
 ## Hardware requirements
 
+### Minimum hardware requirements
+
 Minimum hardware requirements for development include the following:
 
-+ Quad-core Intel or AMD processor, 2.5 GHz or faster
-+ 8 GB RAM
-+ DirectX 12, Vulkan-compatible, or Metal-compatible video card with 2 GB VRAM minimum
-  + 4 GB VRAM recommended
-  + NVIDIA GeForce GTX 1060 or better, driver version 471.11 or later recommended
-  + AMD Radeon Pro 560 or better
-  + Intel HD 630 or better
++ Quad-core 64-bit x86-capable Intel or AMD processor, 2.5 GHz, with the SSE 4.1 SIMD instruction set
++ 16 GB RAM (8GB RAM may be acceptable if you limit the number of processes used for compilation)
++ DirectX 12, Vulkan-compatible, or Metal-compatible video card
+  + 2 GB VRAM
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm (approved):
Looks good to me

micronAMZN (approved):

content/docs/welcome-guide/requirements.md
```diff
@@ -6,30 +6,36 @@ weight: 300
...
+| CPU | Quad-core (4 cores) 64-bit x86-capable Intel or AMD processor, 2.5 GHz, with the SSE 4.1 SIMD instruction set |
+| RAM | 16 GB RAM (8GB RAM may be acceptable if you limit the number of processes used for compilation) |
+| GPU | - DirectX 12 or Vulkan-compatible video card <br> - 2 GB VRAM <br> - Shader Model 6.2 (or Shader Model 6.3 to use raytracing features) <br> - NVIDIA GeForce GTX 1060, driver version 471.11 or later recommended _or_ <br> - AMD Radeon Pro 560 _or_ <br> - Intel HD 630 <br> |
+| Hard Disk | 40 GB (using the [pre-built installer](https://o3debinaries.org/download/windows.html)) _or_ 100+ GB (depending on project configuration) of free disk space |
+
+### Recommended hardware specification
+
+| Hardware Component | Recommended Specification |
+| - | - |
+| CPU | Hexa-core (6 cores) 64-bit x86-capable Intel or AMD processor, 2.5 GHz, supports the SSE 4.1 SIMD instruction set |
```
lmbr-pip, line 34: Do we need to add "or better" or "or faster" here? While its implied, as written, this can be interpreted as you need a machine that matches this specification.

FiniteStateGit: Do you suggest we add that phrasing to each component of the recommended specs and leave it out of minimum specs?

micronAMZN: Maybe classify the two sets of specs as something  like "Minimum Required" and "Minimum Recommended".

byrcolin:
Sig/Platform: I agree with this, looks good.
Should we change "Note requirement of 2GB per core in addition to existing system memory needs required to compile from source." to "Note requirement of 2GB per thread in addition to existing system memory needs required to compile and link from source."
Also should we add a line about mitigation of build problems due to lack of memory per thread like showing them how to limit the number of concurrent compilation units or concurrent links to be under the recommended amount per thread?

byrcolin (approved):

PR #2087, opened 2022-11-15
Website versioning

willihay:
## Change summary

Includes:

- Version switcher dropdown.
- Version warning banner.
- Preview info banner.
- Update site.BaseURL to www.o3de.org.

### Submission Checklist:

@netlify /docs/user-guide

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -25,6 +25,15 @@ contentDir = "/content/"
 docsDir = "/docs/"
 sourceRepo = "https://github.com/o3de/o3de"
 
+[params.docs]
+# Docs version data format: ["display name", "baseURL"]
+# Note: A version that has "www.o3de.org" in its baseURL will get " (latest)" appended to its display name in the version selector.
+versions = [
+  ["22.10", "https://www.o3de.org"],
+  ["development", "https://development--o3deorg.netlify.app"],
+  ["22.05", "https://2205--o3deorg.netlify.app"]
+]
+
 [params.logos]
 navbar = "O3DE-Logo-with-color-white.svg"
 footer = "O3DE-Logo-REV-Mono.svg"
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://o3de.org"
+baseURL = "https://www.o3de.org"
 title = "Open 3D Engine"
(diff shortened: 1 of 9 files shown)


willihay:
Implements RFC https://github.com/o3de/sig-docs-community/issues/54

config.toml
```diff
@@ -1,4 +1,4 @@
-baseURL = "https://o3de.org"
+baseURL = "https://www.o3de.org"
```
AMZN-daimini, line 1: I'd be careful with this change and double check it doesn't impact our SEO

config.toml
```diff
@@ -25,6 +25,15 @@ contentDir = "/content/"
 docsDir = "/docs/"
 sourceRepo = "https://github.com/o3de/o3de"
 
+[params.docs]
+# Docs version data format: ["display name", "baseURL"]
+# Note: A version that has "www.o3de.org" in its baseURL will get " (latest)" appended to its display name in the version selector.
+versions = [
+  ["22.10", "https://www.o3de.org"],
+  ["development", "https://development--o3deorg.netlify.app"],
+  ["22.05", "https://2205--o3deorg.netlify.app"]
```
AMZN-daimini, line 34: Is there any way to get subdomains for these? Like `development.o3de.org` and `2205.o3de.org`

config.toml
AMZN-daimini: @sptramer Would you know about this (and the other comment below)?

config.toml
sptramer: There is! We have to have the Linux Foundation set up the DNS records. So far we do not have a contact to do so, even after months of trying.

AMZN-daimini: It would probably be best to just go live with the final subdomains, but I understand that it could mean waiting for a long time on this... so if the consensus is to use the temp ones in the meantime, I'm fine with it.

Question, does this change actually need to be applied on those versions of the site as well? Or do they share the same sass/js?

willihay: This change will need to be merged / applied to every branch that is selectable in the dropdown (i.e. main, development, and 2205) - and eventually stabilization when that's created for the next release.

AMZN-daimini (approved):

chanmosq (approved):

PR #2093, opened 2022-11-16
Add support for TeX MathML input

FiniteStateGit:
## Change summary

- Add MathJax 3.0 to Hugo template enabling TeX and MathML input.
- Include sample TeX input to smoketest.md

### Submission Checklist:

### DIFF
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -5,3 +5,4 @@
 <script src="https://cdn.jsdelivr.net/npm/ua-parser-js@0.7.25/src/ua-parser.js" integrity="sha256-zkQSSL5rBWlvCbIoTPQikbYHhOeeiTJy9YX9LXLvFBk=" crossorigin="anonymous"></script>
 <script src="https://cdn.jsdelivr.net/npm/lunr@2.3.9/lunr.min.js" integrity="sha256-DFDZACuFeAqEKv/7Vnu1Tt5ALa58bcWZegGGFNgET8g=" crossorigin="anonymous"></script>
 <script src="{{ $app.RelPermalink }}" type="module" integrity="{{ $app.Data.Integrity }}"></script>
+<script src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js" integrity="sha384-Wuix6BuhrWbjDBs24bXrjf4ZQ5aFeFWBuKkFekO2t8xFU0iNaLQfp2K6/1Nxveei" crossorigin="anonymous"></script>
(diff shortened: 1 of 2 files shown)


willihay (approved):

layouts/partials/javascript.html
```diff
@@ -5,3 +5,4 @@
 <script src="https://cdn.jsdelivr.net/npm/ua-parser-js@0.7.25/src/ua-parser.js" integrity="sha256-zkQSSL5rBWlvCbIoTPQikbYHhOeeiTJy9YX9LXLvFBk=" crossorigin="anonymous"></script>
 <script src="https://cdn.jsdelivr.net/npm/lunr@2.3.9/lunr.min.js" integrity="sha256-DFDZACuFeAqEKv/7Vnu1Tt5ALa58bcWZegGGFNgET8g=" crossorigin="anonymous"></script>
 <script src="{{ $app.RelPermalink }}" type="module" integrity="{{ $app.Data.Integrity }}"></script>
+<script src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js" integrity="sha384-Wuix6BuhrWbjDBs24bXrjf4ZQ5aFeFWBuKkFekO2t8xFU0iNaLQfp2K6/1Nxveei" crossorigin="anonymous"></script>
```
sptramer, line 8: Is there a reason not to use [KaTeX](https://katex.org/)? This has been the recommended JS renderer for web from both @hultonha and @jeremyong-az .

sptramer (commented):
Will approve after the decision to use KaTeX vs. TeXML is final.

layouts/partials/javascript.html
hultonha: I have no horse in this race and know nothing about KaTeX or TeXML 😅 @jeremyong-az is definitely the expert 😝 I'll defer to his judgement

hultonha (commented):
Change looks great to me but same as @sptramer I'll wait to check if @jeremyong-az as any objections to this before approving.

sptramer (changes requested):
Checked in with @jeremyong-az on Slack, and here's his feedback:

> Katex is far faster and renders better :)
> https://www.jeremyong.com/cpp/machine-learning/2020/10/23/cpp-neural-network-in-a-weekend/ here's an example with katex

So, would recommend changing the TexML to KaTeX. It looks like the rendering may be slightly more complicated, but there is a blog post I found about setting up proper Hugo integration: https://hyperupcall.github.io/blog/posts/render-latex-with-katex-in-hugo-blog/

sptramer (approved):
Mini-conversation with @FiniteStateGit behind the scenes we agreed that fast rendering with KaTeX can be delayed a while due to limitations behind the scenes in Hugo configuration. We're unlikely to have math-heavy pages that need fast rendering support for a bit so an issue for KaTeX investigation/integration will be filed for pick-up later.

PR #2122, opened 2022-11-23
[Docs] Terrain Developer Guide

mbalfour-amzn:
## Change summary

The "Terrain Developer Guide" is a set of documentation pages targeting developers (as opposed to content creators) that want to use the terrain system from code or replace or extend its functionality. The documentation provides an overview to the architecture, design philosophies, usage, and extension of the system.

### Submission Checklist:


content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+* By convention, all gradients currently listen to `OnEntityVisibilityChanged` and use that to enable or disable the gradient data in the Editor. This makes it easy to turn gradients on and off while authoring, though it's important to recognize that they will still be active at runtime regardless of the authoring visibility setting.
+
+## Terrain Components
+
+Each of the terrain components can be replaced with new components that meet the same requirements. For example, a new component could be created that streams down real-world satellite height data as a replacement for the **Terrain Height Gradient List**. It's also possible to combine components together, such as creating a simplified Terrain Spawner component that implements the functionality of the **Terrain Layer Spawner**, **Terrain Height Gradient List**, and **Terrain Surface Gradient List** on one single component.
+
+Any replacement components need to support the following:
+
+* Implement the request bus that matches the component being replaced (`TerrainSpawnerRequestBus`, ``TerrainAreaHeightRequestBus`, TerrainAreaSurfaceRequestBus`, etc).
```
jckand-amzn, line 32: Extra tick here causes this to not format correctly for TerrainAreaHeightRequestBus.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+* Declare that it implements each service that matches the component being replaced (`TerrainAreaService`, `TerrainHeightProviderService`, `TerrainSurfaceProviderService`, etc) in `GetProvidedServices`.
+* Support thread safety.
+* Call `OnCompositionChanged` or `OnCompositionRegionChanged` whenever any data within the component changes.
+* Call the appropriate APIs on the `TerrainSystemServiceRequestBus` to register/unregister/refresh a terrain region.
+
+## Terrain physics integration
+
+The terrain physics support is provided through a divided component system. One half is the `TerrainPhysicsCollider`, which knows about the generic `TerrainDataRequestBus` for terrain data and the generic `HeightfieldProvider` buses for physics communication, and the other half is the `PhysXHeightfieldCollider`, which knows about PhysX and heightfields, but not about terrain. To integrate terrain with a new physics system, a new `*HeightfieldCollider` component would need to be implemented to replace the PhysX version that works with the `HeightfieldProvider` buses and with the new physics system.
```
jckand-amzn, line 40: Does the * before HeightfieldCollider signify something, or is this a typo?

jckand-amzn (approved):

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
mbalfour-amzn: It's a placeholder mark, because the replacement would be "BulletHeightfieldCollider", "JoltHeightfieldCollider", "PhysX51HeightfieldCollider", or whatever. I can see where it should probably get removed though, it might just cause more confusion than value to have it there.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
mbalfour-amzn: Fixed!

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
mbalfour-amzn: Removed the *.

cgalvan (approved):

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/introduction.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+![Example terrain image.](/images/user-guide/visualization/environments/terrain/terrain-example.png)
+
+## What is a "terrain system"?
+
+A "terrain system" is a system that defines the geometry, color, and surface types for a ground surface. More specifically, a terrain system contains:
+
+* Height, color, and surface data mapped to regions of the world
+* Authoring tools and workflows that support data creation / modification and how to map it to regions
```
chanmosq, line 15: Reworded to use a parallel structure
```suggestion
* Authoring tools and workflows that support creating / modifying data and mapping it to regions
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Level components
+
+The terrain system uses three level components:
+
+| Component | Description |
+| - | - |
+| **Terrain World** | Controls the base terrain system. The entire system is enabled or disabled by adding or removing this component. The configuration parameters that affect the entire terrain world exist on this component as well.|
+| **Terrain World Renderer** | Controls the terrain rendering. It requires the **Terrain World** component to exist, and it provides all of the rendering configuration that needs to be consistent across the entire terrain. This component is kept separate from **Terrain World** so that it's possible to define a conceptual terrain without requiring the overhead of rendering for use cases where a visual representation simply isn't necessary. Also, this provides a "replacement touchpoint" for creating different terrain renderers, such as a voxel-based implementation or a raytracing-based implementation.|
```
chanmosq, line 119: - For "...that needs to be...", the way I read this felt like there's emphasis on how the rendering _needs_ to be consistent, which then makes me wonder what that means and why it needs to be. Instead, I believe this sentence intends to indicate simply that the rendering configuration applies consistently throughout the entire terrain. If my assumption is right, see my suggestion below. 
- For the sentence "This component is kept separate...", broke up the sentence because it was getting too long to read easily.
```suggestion
| **Terrain World Renderer** | Controls the terrain rendering. It requires the **Terrain World** component to exist, and it provides all of the rendering configuration, which applies consistently across the entire terrain. This component is separated from **Terrain World** so that it's possible to define a conceptual terrain without requiring the overhead of rendering. This is useful for cases where a visual representation simply isn't necessary. Also, this provides a "replacement touchpoint" for creating different terrain renderers, such as a voxel-based implementation or a raytracing-based implementation.|
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+| **Terrain World Debugger** | Provides helpful debugging features for visualizing aspects of the terrain system. This component is kept separate from **Terrain World** so that it can easily be completely removed from shipping products.|
```
chanmosq, line 120:
```suggestion
| **Terrain World Debugger** | Provides helpful debugging features for visualizing aspects of the terrain system. This component is kept separated from **Terrain World** so that it's easy to completely remove it from shipping products.|
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+| **Terrain Height Gradient List** | (Optional) Defines the height data for the terrain region. |
+| **Terrain Surface Gradient List** | (Optional) Defines the surface type and weight data for the terrain region. |
+
+Auxiliary Components:
+
+The "auxiliary" components only require a Box Shape on their entity, but not a Terrain Layer Spawner, so they can span across multiple terrain regions instead of being tied to a single one. They _can_ still appear on an entity with a Terrain Layer Spawner, it just isn't required.
+
+| Component | Description |
+| - | - |
+| **Terrain Macro Material** | Defines terrain base coloring for all terrain regions that fall within its volume. Macro materials aren't directly attached to single terrain regions so that they can be authored and streamed at different world resolutions than the terrain regions. |
```
chanmosq, line 141:
```suggestion
| **Terrain Macro Material** | Defines terrain base color for all terrain regions that fall within its volume. Macro materials aren't directly attached to single terrain regions so that they can be authored and streamed at different world resolutions than the terrain regions. |
```

chanmosq (commented):
Looks great so far, will continue reviewing. Submitting my first set of comments.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/introduction.md
```diff
@@ -0,0 +1,46 @@
...
+* One or more physics surfaces that can be used to provide a "physical" simulation of the terrain in the virtual world
+
+## Design tenets
+
+The O3DE terrain system was built with specific design tenets in mind. Enhancements and changes to the system should continue to use these tenets as guiding principles:
+
+| | |
+| - | - |
+| **Enjoyable** | Content creators and engineers should be delighted to use the terrain system and excited by the power and flexibility the tools provide.|
+| **Familiar** | The design should be familiar to content creators that use external terrain authoring tools and should follow similar design patterns to other O3DE workflows.|
```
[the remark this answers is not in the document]
mbalfour-amzn: Actually, should this be "as"? 

"In" makes it sound to me like the design patterns need to be followed in the other workflows, not that we're following the patterns contained in the other workflows. I could also go with "contained in" or "used in"

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+2. EBus encourages decoupling. We want each piece of the system to be able to operate independently and only interact through the API contracts defined on the EBus.
+3. EBus enables functional replacements. As long as a component or system meets the EBus contract, it's possible to replace the functionality in discrete chunks.
+
+Queries to the system should be able to run in parallel and only block on data modifications, and data modifications should run serially and block on both data queries and data modifications. The locking pattern is optimized around the assumption that terrain data is queried in a read-only fashion most of the time, and the data is loaded, unloaded, or modified just a small fraction of the time.
+
+## Single system, multiple components
+
+Another key aspect to the terrain system design is that it provides a singleton system as a single entry point for all terrain queries and management of global terrain settings. However, all of the terrain data and components that feed the system are split apart into discrete pieces. There are many reasons for both of these design decisions.
+
+Single terrain system:
```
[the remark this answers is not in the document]
mbalfour-amzn: It's per level. The system gets activated when you add the Terrain World level component to a level. Each level in a project that doesn't contain that component won't have an active terrain system.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
mbalfour-amzn: I went with "coloring" instead of "color" because "color" sounds a little too singular, and "colors" sounds more like a list of colors than an applied texture. I'm open to more suggestions here though. Maybe "Applies a terrain base color texture to all terrain regions that fall within its volume."?

Here are some examples of what the macro material does to the terrain:
level-with-macro-color
level-with-macro-color-up-close

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
mbalfour-amzn: Should it be "is separated" vs "is kept separated"?

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+               |         |         |                |
+               v         v         v                v
+        .------+---------+---------+------.     .--------.
+       |   LmbrCentral (Shape Components)  |   |   Atom   |
+        '---------------------------------'     '--------'
+```
+
+The **Gradient Signal Gem** provides components that map data in the 0-1 range to world positions. The gradient components are used for defining height data and surface weight data. In particular, the **Image Gradient** provides a convenient workflow for importing and using heightmaps and "splat maps" (surface weight maps) that are generated from external terrain authoring tools.
+
+The **Surface Data Gem** provides a way to define conceptual geometric shells with sets of arbitrary tags which can then be referenced from other systems (like Dynamic Vegetation) that operate on generic surfaces. The terrain system uses the Surface Data tags as the naming system for terrain surface types. It also implements a Surface Data component to make the terrain show up as a surface within the Surface Data system. This enables the Dynamic Vegetation system to treat terrain as a "plantable surface".
```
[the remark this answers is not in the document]
mbalfour-amzn: Would "conceptual geometric volumes" make more sense?

content/docs/user-guide/visualization/environments/terrain/_index.md
```diff
@@ -0,0 +1,24 @@
+---
+linkTitle: Terrain
+title: Terrain
+description: Create terrain through a combination of proceduralism and authored content.
+weight: 100
+toc: true
+---
+
+The terrain system uses a series of components to enable terrain authoring, simulation, and visualization for levels of aribtrary size. To use the terrain system, you must enable the Terrain Gem for your project.
+
```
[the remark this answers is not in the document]
mbalfour-amzn: I'll also format the information below into a table to make it a little bit more aesthetically pleasing with the header there.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
```diff
@@ -0,0 +1,143 @@
...
+  * Regions can be dynamically loaded and unloaded individually, making it easy to implement coarse-grained terrain streaming.
+* Each region can have different subsets of features.
+  * Regions where gameplay occurs can spawn terrain physics representations, but distant terrain regions that are for visuals only don't need to have physics enabled.
+  * Aspects of terrain rendering can be disabled for server-side or machine learning simulations that don't require a visual representation of terrain.
+* Different components can have different input data resolutions. The terrain system defines a single _output_ terrain grid and resolution, but the input data resolutions can be varied to optimize the data size and complexity based on the need for that region. Even within a region, two different data providers (such as height and surface type) can use two different resolutions based on what's appropriate for that type of data.
+* The functionality can be replaced at a component-by-component level.
+  * The PhysX Heightfield component could be replaced with one that connects to another physics system such as Jolt or Bullet.
```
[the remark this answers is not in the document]
mbalfour-amzn: Those links are correct, but I don't see any obvious trademark information. Would it be correct to refer to them as "Jolt Physics" and "Bullet Physics SDK", since that's how they are referred to on their main github pages?

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
chanmosq: Yes I think "a terrain base color texture" is perfect. Texture is a clue to the reader what kind of asset is used.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
chanmosq: Ah yes, I think I meant to remove "kept" as I did in other comments. Thanks for catching!

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
mbalfour-amzn: I'll change the first one to "it provides all of the rendering configuration settings that apply globally to the entire terrain." The other part looks good to me.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/introduction.md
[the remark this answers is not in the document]
chanmosq: Ah yes, that would make it more clear. 
Both your suggestions are fine, or even "as in" is okay. 
If you prefer to make changes using the batch suggestion on the PR, feel free to update my comment here to match the wording you choose.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/architecture.md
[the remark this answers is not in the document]
chanmosq: Yes! Using the name they use to refer to themselves is a safe bet.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+   0.00   0.25   0.50   0.75   1.00
+```
+
+| Height Query Resolution | Result | Description |
+| - | - | - |
+| 1.0 meters | 40 | Querying at 0.75 will fetch the underlying data at 0.00 (40) and 1.00 (40) and interpolate at 75%. |
+| 0.5 meters | 30 | Querying at 0.75 will fetch the underlying data at 0.50 (20) and 1.00 (40) and interpolate at 50%. |
+| 0.25 meters | 10 | Querying at 0.75 will fetch the underlying data at 0.75 (10) and 1.00 (40) and interpolate at 0%. |
+
+The `Default` Sampler setting is the one that should be used by most systems when querying. However, there are two other sampling methods that are available for specialized use.
```
chanmosq, line 55:
```suggestion
The `Default` Sampler setting should be used by most systems when querying. However, there are two other sampling methods that are available for specialized use.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+
+The `Clamp` Sampler setting will take the requested query position and 'clamp' it down to the terrain grid before requesting the value. Where the `Bilinear` sampler produces an infinite range of interpolated values for all the positions between grid points, the `Clamp` sampler will always produce the same value for every position between grid points. Because this only fetches one data value, not four, it can run up to 4x faster than the `Default` sampler, but the values returned will only match the rendering and physics representations when queried directly on the grid points. For that reason, this setting should only be used when querying directly on terrain grid points.
```
chanmosq, line 57: Small suggestion to recall that there are "four" positions here because it helps quickly answer "why is it 4x faster?"

```suggestion
The `Clamp` Sampler setting takes the requested query position and "clamps" it down to the terrain grid before requesting the value. Whereas the `Bilinear` sampler produces an infinite range of interpolated values for all four positions between the grid points, the `Clamp` sampler always produces the same value for each position. Because this only fetches one data value, not four, it can run up to 4x faster than the `Default` sampler, but the returned values only match the rendering and physics representations when they are queried directly on the grid points. For that reason, this setting should only be used when querying directly on terrain grid points.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+
+The `Exact` Sampler setting will take the requested query position and pass it directly down to the underlying data source, ignoring the terrain grids entirely. Similar to `Clamp`, this is 4x faster than `Default`, but will only produce matching values when queried directly on the grid points. This setting should also only be used when querying directly on terrain grid points, but it may serve some alternate purpose if there's ever a reason to query the underlying terrain data directly without regards to the terrain grid.
```
chanmosq, line 59: My suggestion reworks your last sentence -- I wanted to remove "but" because I think this is an "and" rather than a "but" situation. I also attempt to streamline the last part, such as removing unnecessary words. 
```suggestion
The `Exact` Sampler setting takes the requested query position and passes it directly down to the underlying data source, ignoring the terrain grids entirely. Similar to `Clamp`, this is 4x faster than `Default`, but only produces matching values when it's queried directly on the grid points. You should only use this setting when you are querying directly on the terrain grid points. Alternatively, `Exact` may be useful if you want to directly query the underlying terrain data, without regards to the terrain grid.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+
+### The `terrainExists` flag
+
+Because terrain data only exists within arbitrary authored regions, it's possible to query the terrain system for points where the terrain does not exist. The terrain is defined as existing as long as height data exists. When querying for terrain data, this flag should always be checked for each query result to ensure that the data actually represents valid terrain.
+
+### Query optimizations
+
+Here are a few tips for getting the best performance out of the terrain query APIs:
+
+* The APIs are designed to let you query for different combinations of height, normal, and surface data. The more data you query, the more expensive the query will be, so try to limit queries to fetch the minimum amount of data that you intend to use. When you need multiple types of data for the same position, the APIs that let you fetch multiple types at once will generally be faster than calling the individual APIs separately. For example, `GetSurfacePoint` will be faster than calling `GetHeight`, `GetNormal`, and `GetSurfaceWeights` individually.
```
chanmosq, line 69:
```suggestion
* The APIs are designed to let you query for different combinations of height, normal, and surface data. The more data you query, the more expensive the query is, so try to limit queries to fetch the minimum amount of data that you intend to use. When you need multiple types of data for the same position, the APIs that let you fetch multiple types at once are generally faster than calling the individual APIs separately. For example, `GetSurfacePoint` is faster than calling `GetHeight`, `GetNormal`, and `GetSurfaceWeights` individually.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+* When querying for multiple input positions, the bulk APIs (`QueryList`, `QueryRegion`) are generally 80-90% faster than calling the individual APIs (`GetHeight`, etc).
+* `QueryRegion` / `QueryRegionAsync` are the fastest query APIs because they don't require the additional step of building up an input list of positions before calling them.
+* The queries can be made up to 4x faster by using the `Exact` or `Clamp` sampler choices, as long as you take care to align the input positions or input region exactly to the terrain grid. Misalignment will produce inaccurate results.
```
chanmosq, line 72:
```suggestion
* The queries can be up to 4x faster by using the `Exact` or `Clamp` sampler choices, as long as you take care to align the input positions or input region exactly to the terrain grid. Misalignment produces inaccurate results.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+* The `TerrainAreaExistsInBounds` provides a quick way to determine if terrain exists at _all_ within an AABB. This can be used to circumvent the more expensive per-point queries entirely if the region your system cares about doesn't contain any terrain.
```
chanmosq, line 73:
```suggestion
* The `TerrainAreaExistsInBounds` provides a quick way to determine if terrain exists within an AABB. This can be used to entirely circumvent the more expensive per-point queries if the region that your system cares about doesn't contain any terrain.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Raycast queries
+
+The `TerrainDataRequestBus` provides a `GetClosestIntersection` method to perform a single raycast against the terrain. Typically, collision raycasts are expected to be performed through a physics system, but this method is available as an alternative for when physics raycasts aren't appropriate. For example, it's used within the Editor itself for placing objects on the terrain.
+
+If your system is always raycasting straight down to get terrain information directly beneath a location, the query APIs will produce faster results than the raycast API.
```
chanmosq, line 79: I'm assuming the query APIs refers to the ones mentioned earlier on this page, and the raycast API refers to this `GetClosestIntersection` function? If so, I suggest adding that clarity here. I think it starts to get confusing which API you're talking about because this section is called "raycast queries"

```suggestion
If your system often raycasts straight down to get terrain information directly beneath a location, the query APIs that are mentioned earlier produces faster results than this raycast API.
```

chanmosq (commented):
Reviewed `system-usage`

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+Gradient components are the underlying data source for height data and surface weight data. New gradient component types can be created as-needed to provide data from different sources, such as streaming satellite data, or to modify existing gradient data, such as adding complex erosion filters.
+
+New gradient components need to support the following at a minimum:
+
+* Implement the `GradientRequestBus` to generically return a floating-point output value for an input world position.
+* Declare that it implements the `GradientService` in `GetProvidedServices`.
+* Support thread safety. Ideally, this will mean using a `shared_mutex` with a `shared_lock` in the query APIs and a `unique_lock` for any places that change the underlying data used in the query API methods. This will let multiple queries run in parallel while safely blocking on data changes.
+* Call `DependencyNotificationBus::Events::OnCompositionChanged` or `DependencyNotificationBus::Events::OnCompositionRegionChanged` whenever any data within the gradient changes. This notifies everything "above" the gradient that it will need to refresh itself.
+
+Ideally, new gradient components would also support the following:
```
chanmosq, line 20: I suggest using an alternative to "ideally" to circumvent explaining why this is ideal, which readers might wonder. "should" is the more appropriate form here. 
```suggestion
In addition, new gradient components should also support the following:
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+
+* Use the **Shape** components to define how the gradient data should map into world space.
```
chanmosq, line 22:
```suggestion
* Use the **Shape** components to define how the gradient data maps into world space.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+* Support the **Gradient Transform Modifier** component for performing arbitrary transforms to the gradient data.
+* By convention, all gradients currently listen to `OnEntityVisibilityChanged` and use that to enable or disable the gradient data in the Editor. This makes it easy to turn gradients on and off while authoring, though it's important to recognize that they will still be active at runtime regardless of the authoring visibility setting.
```
chanmosq, line 24:
```suggestion
* By convention, all gradients currently listen to `OnEntityVisibilityChanged` and use that to enable or disable the gradient data in the Editor. This makes it easy to turn gradients on and off while authoring, though it's important to recognize that they are still active at runtime regardless of the authoring visibility setting.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Terrain Components
+
+Each of the terrain components can be replaced with new components that meet the same requirements. For example, a new component could be created that streams down real-world satellite height data as a replacement for the **Terrain Height Gradient List**. It's also possible to combine components together, such as creating a simplified Terrain Spawner component that implements the functionality of the **Terrain Layer Spawner**, **Terrain Height Gradient List**, and **Terrain Surface Gradient List** on one single component.
```
chanmosq, line 28:
```suggestion
Each of the terrain components can be replaced with new components that meet the same requirements. For example, a new component that streams down real-world satellite height data can replace the **Terrain Height Gradient List**. It's also possible to combine components together, such as a simplified Terrain Spawner component that implements the functionality of the **Terrain Layer Spawner**, **Terrain Height Gradient List**, and **Terrain Surface Gradient List** in one single component.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+
+Any replacement components need to support the following:
+
+* Implement the request bus that matches the component being replaced (`TerrainSpawnerRequestBus`, `TerrainAreaHeightRequestBus`, `TerrainAreaSurfaceRequestBus`, etc).
+* Declare that it implements each service that matches the component being replaced (`TerrainAreaService`, `TerrainHeightProviderService`, `TerrainSurfaceProviderService`, etc) in `GetProvidedServices`.
```
chanmosq, line 33: - Nit, this is mostly personal preference and standard docs style. I do prefer "For example", or similar, as makes it clear what the info your listing is for. 
  Either way is okay for me, but I strongly suggest moving the examples in the second bullet to the end of the sentence. Since the examples have long names, it interrupts the sentence. 
- Also, for lists, keep your punctuation consistent. Either no period or a period for all of them. 
```suggestion
* Implement the request bus that matches the component being replaced. For example, `TerrainSpawnerRequestBus`, `TerrainAreaHeightRequestBus`, `TerrainAreaSurfaceRequestBus`, etc.
* Declare that it implements each service that matches the component being replaced in `GetProvidedServices`. For example, TerrainAreaService`, `TerrainHeightProviderService`, `TerrainSurfaceProviderService`, etc.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+* Support thread safety.
+* Call `OnCompositionChanged` or `OnCompositionRegionChanged` whenever any data within the component changes.
+* Call the appropriate APIs on the `TerrainSystemServiceRequestBus` to register/unregister/refresh a terrain region.
+
+## Terrain physics integration
+
+The terrain physics support is provided through a divided component system. One half is the `TerrainPhysicsCollider`, which knows about the generic `TerrainDataRequestBus` for terrain data and the generic `HeightfieldProvider` buses for physics communication, and the other half is the `PhysXHeightfieldCollider`, which knows about PhysX and heightfields, but not about terrain. To integrate terrain with a new physics system, a new `HeightfieldCollider` component would need to be implemented to replace the PhysX version that works with the `HeightfieldProvider` buses and with the new physics system.
```
chanmosq, line 40:
```suggestion
The terrain physics support is provided through a divided component system. One half is the `TerrainPhysicsCollider`, which knows about the generic `TerrainDataRequestBus` for terrain data and the generic `HeightfieldProvider` buses for physics communication. The other half is the `PhysXHeightfieldCollider`, which knows about PhysX and heightfields, but not about terrain. To integrate terrain with a new physics system, a new `HeightfieldCollider` component needs to be implemented to replace the PhysX version that works with the `HeightfieldProvider` buses and with the new physics system.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
mbalfour-amzn: This might need better clarification, "four" isn't a correct change here. 

Suppose you have four grid points like this where the heights are 5 at X=1, and 6 at X=2:
```
2 +  *5 *6
  |
1 +  *5 *6
  |    
0 +--+--+
   0  1  2
```

What this section is trying to say is that you can query an infinite number of positions between those grid points and get back an infinite number of results. For example, a query of (1.5, 1) gives you a height of 5.5. a query of (1.5001, 1) gives you a height of 5.5001. And so on. That's what "all the positions" is supposed to mean in this context.

In contrast, when using using 'Clamp', a query of (1.5, 1) gives you a height of 5. A query of (1.5001, 1) gives you a height of 5. A query of (1.999, 1) gives you a height of 5. But a query of (2, 1) gives you a height of 6.

Open to suggestions on how to make this more clear...

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Terrain renderer
+
+The entire terrain renderer could be replaced by creating new rendering level components, rendering entity components, feature processors, and shaders. Replacing the renderer will have no effect on the base terrain system or the terrain physics integrations. This undertaking isn't for the faint of heart, as a high-performance terrain renderer can be difficult to write.
```
chanmosq, line 44: I don't think the last sentence is well suited for a developer guide style (perhaps a blog style). While it is true, it may come off as discouraging to some. 
```suggestion
The entire terrain renderer can be replaced by creating new rendering level components, rendering entity components, feature processors, and shaders. Replacing the renderer doesn't affect the base terrain system or the terrain physics integrations.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Terrain system
+
+The underlying terrain system could be replaced entirely as well, as long as it implements the `TerrainDataRequestBus` and `TerrainSystemServiceRequestBus` and it uses the `TerrainDataNotificationBus` to send change notifications. There's no obvious reason to replace the terrain system itself in isolation though, as it mostly just routes data requests and notifications.
```
chanmosq, line 48:
```suggestion
The underlying terrain system can be replaced entirely as well, as long as it implements the `TerrainDataRequestBus` and `TerrainSystemServiceRequestBus`, and it uses the `TerrainDataNotificationBus` to send change notifications. There's no obvious reason to replace the terrain system itself in isolation though, as it mostly just routes data requests and notifications.
```

chanmosq (commented):
Submitting review for system-extensions.md

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+The API extensions would likely consist of redefining the current APIs to provide the _highest elevation_ terrain data that appears below the given query input positions, and then adding more APIs to provide _all_ the terrain data that exists at or below the given query positions.
+
+### GPU calculation support
+
+The terrain system is built on top of the gradient components, which perform all of their calculations CPU-side. However, the types of calculations being performed are ideal for GPU-side calculation. An overhaul or replacement of the gradient system that supports GPU-side calculations could provide orders of magnitude improvements to the terrain system. The primary caveats are that the data is still needed CPU-side to feed to physics, generalized raycasts, and other systems, and that it should be possible to still use the terrain system on devices with limited or no GPU capabilities (low-end phones, headless servers, etc).
```
chanmosq, line 23: For various reasons, I suggest using a simpler adjective, rather than orders of magnitude. The primary reason is that it's long and I think breaks the reading flow. 
```suggestion
The terrain system is built on top of the gradient components, which perform all of their calculations on the CPU-side. However, these calculations can be computed more optimally on the GPU-side. An overhaul or replacement of the gradient system that supports GPU-side calculations can provide significant improvements to the terrain system's performance. The primary caveats are that the data is still needed on the CPU-side to feed to physics, generalized raycasts, and other systems, and that it should be possible to still use the terrain system on devices with limited or no GPU capabilities (low-end phones, headless servers, etc).
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Additional features
+
+This category of improvements should be possible to incrementally add into the system without fundamentally altering anything about how the system currently works.
```
chanmosq, line 27:
```suggestion
This category of improvements may be incrementally added to the system without fundamentally altering anything about how the system currently works.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Triangle split direction
+
+The terrain currently always triangulates each quad in the same uniform direction. This could be replaced with heuristics for choosing the best direction for each quad. This could even be exposed as multiple heuristics - same direction for every quad, alternating directions per quad, best choice per quad based on vertex heights. This decision should be controlled at the terrain system level, so that the information can be provided consistently to the terrain debugger, terrain physics, terrain rendering, terrain raycasting, etc.
```
chanmosq, line 31:
```suggestion
Currently, the terrain always triangulates each quad in the same uniform direction. This can be replaced with heuristics for choosing the best direction for each quad. This can even be exposed as multiple heuristics: same direction for every quad, alternating directions per quad, or best choice per quad based on vertex heights. This decision should be controlled at the terrain system level, so that the information can be provided consistently to the terrain debugger, terrain physics, terrain rendering, terrain raycasting, etc.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Improved shape to height workflows
+
+It's currently somewhat non-intuitive to create terrain heights directly from primitive shape components. The best options are to use either the Shape Falloff Gradient or the Surface Altitude Gradient, but they're both currently problematic. The Shape Falloff Gradient creates falloff based on distance from the bottom of a box, not just simply the position of the shape within a box, so it's hard to control. The Surface Altitude Gradient doesn't support falloff and the auto-refresh doesn't work. Either of these components could be improved, or a new component could be added, to make it easier to just place a shape with falloff into the world and turn it into a height gradient.
```
chanmosq, line 35:
```suggestion
Currently, it's somewhat non-intuitive to create terrain heights directly from primitive shape components. The best options are to use either the Shape Falloff Gradient or the Surface Altitude Gradient, but they're both problematic at this moment. The Shape Falloff Gradient creates falloff based on distance from the bottom of a box, not simply the position of the shape within a box, making it hard to control. The Surface Altitude Gradient doesn't support falloff and the auto-refresh doesn't work. Either of these components can be improved or a new component can be added, to make it easier to place a shape with falloff into the world and turn it into a height gradient.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Masking / blending
+
+Add controls to the Terrain Layer Spawner and the Terrain Macro Material to allow each one to mask and blend instead of completely replacing the data. This would make it much easier to create small terrain "stamps" for things like craters.
```
chanmosq, line 39:
```suggestion
Add controls to the Terrain Layer Spawner and the Terrain Macro Material to allow each one to mask and blend instead of completely replacing the data. This will make it easier to create small terrain "stamps" for things like craters.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+### First-class "hole" support
+
+Holes can be created right now by authoring a high-priority terrain layer spawner with no ground plane for whatever size is desired. However, it would be nice to have additional authoring controls to make it possible to put holes directly into gradient data, possibly either through a second alpha channel, or through a separate gradient on the Terrain Height Gradient List component.
```
chanmosq, line 43:
```suggestion
Holes can be created right now by authoring a high-priority terrain layer spawner with no ground plane for whatever size is desired. However, it will be nice to have additional authoring controls to make it possible to put holes directly into gradient data, possibly either through a second alpha channel, or through a separate gradient on the Terrain Height Gradient List component.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Improved terrain texturing
+
+There are a number of features that would be nice to add to the terrain texturing:
```
chanmosq, line 47:
```suggestion
There are a number of features that will be nice to add to the terrain texturing:
```

chanmosq (commented):
Reviewed future-work.md

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
mbalfour-amzn: Hmm, how about this instead?

```suggestion
The `Exact` Sampler setting takes the requested query position and passes it directly down to the underlying data source, ignoring the terrain grids entirely. Similar to `Clamp`, this is 4x faster than `Default`, but only produces matching values when it's queried directly on the grid points. You should not use it to query positions outside of the grid points, since the data may not align with the rendering or physics representations. However, the feature is available for those who want it.
```

The basic points are that the data from `Exact` is useful when queried directly on the grid points, _probably_ isn't useful anywhere else, but _maybe_ someone will have a neat way to use it and we don't want to prevent them from doing it.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
mbalfour-amzn: Hm, after re-reading this block, how about this instead:

```suggestion
{{< tip >}}
`GetClosestIntersection` provides generalized raycasts in any direction, but if your system only raycasts straight down to get terrain information directly beneath a location, use one of the other query APIs such as `GetSurfacePoint` or `QueryList` instead. They produce faster results.
{{< /tip >}}
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
mbalfour-amzn: We're also not supposed to use Latin abbreviations, so how's this instead?

```suggestion
* Implement the request bus that matches the component being replaced. For example, `TerrainSpawnerRequestBus`, `TerrainAreaHeightRequestBus`, or `TerrainAreaSurfaceRequestBus`.
* Declare that it implements each service that matches the component being replaced in `GetProvidedServices`. For example, `TerrainAreaService`, `TerrainHeightProviderService`, or `TerrainSurfaceProviderService`.
```

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+There are a number of improvements that could potentially be made to the terrain system for anyone interested in taking the system further.
+
+## Large-scale improvements
+
+This category of improvements requires architecting new solutions to provide extensive new capabilities.
+
+### 3D geometry support
+
+Instead of a heightfield-based approach, the APIs and renderer could be extended to use a full 3D approach, whether it's through voxels, displacement maps, mesh ingestion, raytracing, or some other technology. This would enable the system to support caves, arches, cliff geometry, and other vertical or overlapping terrain features that simply can't be supported through a pure heightfield approach.
```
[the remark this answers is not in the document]
mbalfour-amzn: For my edification, can you explain the use of "will" instead of "would" here? Since this is a hypothetical scenario that hasn't happened yet, isn't "would" the correct choice?

i.e. "If I extend it, it would support caves" vs "If I extend it, it will support caves".

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
+---
+linktitle: Future Work
+title: Future Work
+description: Possible future improvements to the terrain system.
+weight: 500
+toc: true
+---
+
+There are a number of improvements that could potentially be made to the terrain system for anyone interested in taking the system further.
```
mbalfour-amzn, line 9:
```suggestion
There are a number of improvements that can potentially be made to the terrain system for anyone interested in taking the system further. You can reach out to the O3DE community to talk through any terrain enhancements or ideas on the #sig-content channel on [Discord ](https://discord.com/invite/o3de). We are happy to help!
```

mbalfour-amzn: How's that sound?

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Large-scale improvements
+
+This category of improvements requires architecting new solutions to provide extensive new capabilities.
+
+### 3D geometry support
+
+Instead of a heightfield-based approach, the APIs and renderer could be extended to use a full 3D approach, whether it's through voxels, displacement maps, mesh ingestion, raytracing, or some other technology. This would enable the system to support caves, arches, cliff geometry, and other vertical or overlapping terrain features that simply can't be supported through a pure heightfield approach.
+
+The API extensions would likely consist of redefining the current APIs to provide the _highest elevation_ terrain data that appears below the given query input positions, and then adding more APIs to provide _all_ the terrain data that exists at or below the given query positions.
```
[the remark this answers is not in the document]
mbalfour-amzn: Are you sure? I think "may" means "possible", but doesn't imply "likely". (The definition of "may" is _literally_ just "expressing possibility.")

If the goal is to remove "would", how about "will likely" instead of "would likely"? This is one place that I can understand the would/will swap, since the addition of "likely" keeps it in the hypothetical instead of the absolute.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
mbalfour-amzn: How about:
```suggestion
The terrain system is built on top of the gradient components, which perform all of their calculations on the CPU. However, a GPU can perform bulk computations like these far more optimally. An overhaul or replacement of the gradient system that supports GPU-side calculations can improve the terrain system's performance by orders of magnitude. The primary caveats are that the data is still needed on the CPU to feed to physics, generalized raycasts, and other systems, and that it should be possible to still use the terrain system on devices with limited or no GPU capabilities (low-end phones, headless servers, etc).
```

The phrase "orders of magnitude" is more specific and objective than "significant" so I'd like to preserve it.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
chanmosq: Yes, sounds good!

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
chanmosq: Yes that sounds clearer! Small note: "generalized" -> "general"

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
chanmosq: Hm, I was hoping my editing on your suggestion would show so you can see it. (I noticed that works in GH issues.) So apologies for lack of transparency there. I made two small changes, changed "talk through" to "talk about", and removed a space after "Discord".

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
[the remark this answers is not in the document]
chanmosq: Oh yes, sorry for the miss. Given the hypothetical or conditional statement, "would" is more accurate.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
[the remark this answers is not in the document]
chanmosq: Yeah, at least compared to "might", "may" indicates greater chance of possibility. I understand that probably isn't intuitive for readers though. "will likely" is perfectly fine

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/future-work.md
chanmosq: I wasn't sure if it was figurative or not, so in that case, agree to keep orders of magnitude, and your edit looks good.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-extensions.md
chanmosq: You're right, we usually don't use latin abbrevs. I know that with this going into the Engine Developer Guide eventually, that will be under a Style Guide that's decided between the technical sigs and sig-docs-community. So I've been reviewing it more lightly on some rules. Not using latin abbrevs can be a style we choose to keep and I'm +1 for that.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
mbalfour-amzn: Discussed offline, we'll keep "generalized" since it has an intentional use in this sentence.

content/docs/user-guide/visualization/environments/terrain/terrain-developer-guide/system-usage.md
mbalfour-amzn: I've reworked this suggestion and included (hopefully) better diagrams. What do you think?

PR #2130, opened 2022-11-29
Make search form always use relative path

cgalvan:
## Change summary

Change the search form to use the local `/search` path instead of only using `o3de.org` or `localhost`. This allows the deployed preview site search to operate on the deployed version, instead of being redirected to `o3de.org/search`.

### Submission Checklist:

Signed-off-by: Chris Galvan <chgalvan@amazon.com>

### DIFF
--- a/layouts/partials/search-results.html
+++ b/layouts/partials/search-results.html
@@ -26,7 +26,7 @@
           // You can customize your searchable fields using any .Page parameters
           "title": "{{ .Title  }}",
           "content": {{ .Content | plainify }}, // Strip out HTML tags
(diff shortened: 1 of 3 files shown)


AMZN-daimini:
The search results also need to be fixed as they always redirect to o3de.org right now

cgalvan:
> The search results also need to be fixed as they always redirect to o3de.org right now

Good call! I've added a fix for that as well.

AMZN-daimini (approved):

sptramer (approved):

PR #2170, opened 2022-12-14
Updated the windows release packaging instructions

cgalvan:
## Change summary

Updated the windows release packaging instructions.
- Made it more obvious that the user must use a project-centric configuration
- Added explanations of monolithic vs. non-monolithic in regards to game only vs. game + editor/tools and why monolithic would be the common case
- Re-arranged the monolithic vs. non-monolithic build sections so that monolithic is the first option
- Removed mentions of the asset bundling step being optional, and added explanation as to why it is important

### Submission Checklist:

Signed-off-by: Chris Galvan <chgalvan@amazon.com>

### DIFF
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -191,7 +191,7 @@ The other option is to release your game plus the **O3DE Editor** and all of its
 
 Building from a source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
 
-To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `install` target and `release` configuration.
+To build your non-monolithic project for release, use CMake to invoke Visual Studio MSBuild. Specify the `install` target and `release` configuration.
 
 ```cmd
 cd C:\MyProject
@@ -242,7 +242,7 @@ The most common workflow is to release your game only, in which case you should 
 
     ```cmd
     cd C:\MyProject
-    cmake --B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\MyProject\MyProjectGameLayout
+    cmake --B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\MyProject\MyProjectGameLayout -DLY_MONOLITHIC_GAME=1
     ```
 
 1. Finally, invoke the CMake build wrapper to build the project game release layout using the `install` target.
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -93,7 +93,9 @@ To process your project's assets, do one of the following:
 
 - **Run Asset Processor.** Do this if you already built Asset Processor and prefer to use a GUI.
 
-    1. Run `AssetProcessor.exe` from the `<project>/build/windows/bin/profile` directory.
+    1. Run `AssetProcessor.exe` from the build directory based on your engine type.
+        - `<project>/build/windows/bin/profile` -- For source engine.
+        - `<engine>/build/windows/bin/profile` -- For pre-built SDK engine.
 
         {{< image-width "/images/user-guide/packaging/windows-release-build/asset-processor.png" "700" "An image of O3DE Asset Processor." >}}
 
@@ -101,7 +103,9 @@ To process your project's assets, do one of the following:
 
 - **Run Asset Processor Batch.** Do this if you already built Asset Processor and prefer to use a CLI.
 
-    1. Run `AssetProcessorBatch.exe` from the `<project>\build\windows\bin\profile` directory.
+    1. Run `AssetProcessorBatch.exe` from the build directory based on your engine type.
+        - `<project>/build/windows/bin/profile` -- For source engine.
+        - `<engine>/build/windows/bin/profile` -- For pre-built SDK engine.
 
         {{< image-width "/images/user-guide/packaging/windows-release-build/asset-processor-batch.png" "700" "An image of O3DE Asset Processor Batch." >}}
 
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -9,7 +9,7 @@ weight: 200
 This tutorial guides you through the process of creating an **Open 3D Engine (O3DE)** *project game release layout* for Windows computers. A project game release layout is a directory structure that contains the **Game Launcher** and the bundled assets needed to run the Game Launcher outside of the developer environment. You create a project game release layout when you build your project for release, known as a *release build*.
 
 {{< important >}}
-In order to create a project game release layout, you must have O3DE set up from source, not just the installer. Refer to [Setting up O3DE from GitHub](/docs/welcome-guide/setup/setup-from-github/). 
+In order to create a game only release (monolithic), you must have O3DE set up from source, not just the installer. Refer to [Setting up O3DE from GitHub](/docs/welcome-guide/setup/setup-from-github/).
 Also, you must use a project-centric configuration, which is detailed further [below](#create-a-project-game-release-layout).
 {{< /important >}}
 
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -224,19 +224,28 @@ Now, you will choose a build type for your project game release layout.
 
 The most common workflow is to release your game only, in which case you should choose the **monolithic** build. This will build only your game/server launcher and any necessary dependencies.
 
-1. Use CMake to create a Visual Studio project for an engine that supports monolithic projects. Specify a new CMake build directory (`build\windows_mono`) in your pre-built SDK engine directory, separate from your non-monolithic build directory. Specify a monolithic build by enabling the `-D` option, `LY_MONOLITHIC_GAME`.
+1. First, we need to build the monolithic artifacts in your SDK engine layout. Specify a new CMake build directory (`build\windows_mono`) in your pre-built SDK engine directory, separate from your non-monolithic build directory. Specify a monolithic build by enabling the `-D` option, `LY_MONOLITHIC_GAME`.
 
     ```cmd
     cd C:\o3de
     cmake --preset windows-mono-default -DLY_VERSION_ENGINE_NAME=o3de-install
     ```
 
-2. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
+1. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
 
     ```cmd
     cd C:\o3de
-    cmake --build --preset windows-mono-install  --config release
+    cmake --build --preset windows-mono-install --config release
     ```
+
+1. Now that the monolithic build artifacts have been added to the SDK engine layout, your project can configure against that layout to run a monolithic build. Switch to your project's source directory and use CMake to create a Visual Studio project for your project that builds against the monolithic engine.
+
+    ```cmd
+    cd C:\MyProject
+    cmake --B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\MyProject\MyProjectGameLayout
+    ```
+
+1. Finally, invoke the CMake build wrapper to build the project game release layout using the `install` target.
 
     This command generates a Game Launcher and dependent `.lib` files. It also bundles your project's product assets that are located in `<project>\Cache\pc` into an `engine.pak` file.
 
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -166,9 +166,7 @@ The most common workflow is to release your game only, in which case you should 
     cmake -B build\windows_mono -S . -G "Visual Studio 16" -DLY_3RDPARTY_PATH=C:\o3de-packages -DLY_MONOLITHIC_GAME=1
     ```
 
-    {{< note >}}
-Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
-    {{< /note >}}
+    **Note:** Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
 
 1. To build your monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `INSTALL` target and `release` configuration.
 
@@ -256,9 +254,7 @@ A pre-built SDK engine supports non-monolithic projects by default. As detailed 
     cmake -B build/windows -G "Visual Studio 16" -DLY_3RDPARTY_PATH=C:\o3de-packages -DLY_VERSION_ENGINE_NAME=o3de-install -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_PROJECTS=C:\o3de-projects\MyProject
     ```
 
-    {{< note >}}
-Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
-    {{< /note >}}
+    **Note:** Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
 
 1. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
 
--- a/content/docs/user-guide/packaging/windows-release-builds.md
+++ b/content/docs/user-guide/packaging/windows-release-builds.md
@@ -187,7 +187,7 @@ Your release build is now complete! Next, you need to [bundle your content](#bun
 
 The other option is to release your game plus the **O3DE Editor** and all of its tools, which is is a **non-monolithic** build. This is not recommended since the size of all these tools combined will be much larger than the **monolithic** build.
 
-A source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
+Building from a source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
 
 To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `INSTALL` target and `release` configuration.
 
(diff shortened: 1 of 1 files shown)


content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -194,6 +181,27 @@ Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Stud
 
 The result is a project game release layout in the install directory that's located at `<install>\bin\Windows\release\Monolithic`. In your project game release layout, your `engine.pak` file is located in `Cache\pc`.
 
+Your release build is now complete! Next, you need to [bundle your content](#bundle-content).
+
+#### Option 2: Non-monolithic (Game + Editor/Tools)
+
+The other option is to release your game plus the **O3DE Editor** and all of its tools, which is is a **non-monolithic** build. This is not recommended since the size of all these tools combined will be much larger than the **monolithic** build.
+
+A source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
```
lmbr-pip, line 190: nit: Seems like a missing word `A source engine (build)`?

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -194,6 +181,27 @@ Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Stud
...
+
+To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `INSTALL` target and `release` configuration.
+
+```cmd
+cd C:\MyProject
+cmake --build build\windows --target INSTALL --config release
+```
+
+You can specify a particular non-monolithic build by appending the option `-DLY_MONOLITHIC_GAME=0`. This command generates O3DE tools (such as Editor, Asset Processor, and Game Launcher) and dependent `.dll` files. It also bundles your project's product assets that are located in `<project>\Cache\pc` into an `engine.pak` file.
```
lmbr-pip, line 199: nit: Mention ServerLauncher as we will generate it via this path I believe?

lmbr-pip (approved):

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -92,15 +93,15 @@ To process your project's assets, do one of the following:
...
+    1. Run `AssetProcessor.exe` from the `<project>/build/windows/bin/profile` directory.
 
         {{< image-width "/images/user-guide/packaging/windows-release-build/asset-processor.png" "700" "An image of O3DE Asset Processor." >}}
 
     1. Wait for all jobs to complete. You can view the status of each asset in the **Asset Status** pane and the details of each asset in the **Event Log Details** pane.
 
 - **Run Asset Processor Batch.** Do this if you already built Asset Processor and prefer to use a CLI.
 
-    1. Run `AssetProcessorBatch.exe` from the `<engine>\build\windows\bin\profile` directory.
+    1. Run `AssetProcessorBatch.exe` from the `<project>\build\windows\bin\profile` directory.
```
lumberyard-employee-dm, line 104: Uh.. about this.
If the user is using a SDK engine, then there Asset Processor would be in the SDK layout directory as part of the engine.

The engine tools like the Asset Processor is placed in the build directory of the project when a project-centric source-engine workflow is used.
When a project-centric SDK engine workflow is used, the AssetProcessor(Batch) has already been built and the one in the SDK layout folder is what gets used.

We should mention the differences in these details in the "Source Engine" and "Pre-built SDK engine" sections which are under heading 3 "###" below

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -150,27 +151,13 @@ In this step, you use CMake to build either a non-monolithic or a monolithic pro
...
-cmake --build build\windows --target INSTALL --config release 
-```
-
-You can specify a particular non-monolithic build by appending the option `-DLY_MONOLITHIC_GAME=0`. This command generates O3DE tools (such as Editor, Asset Processor, and Game Launcher) and dependent `.dll` files. It also bundles your project's product assets that are located in `<project>\Cache\pc` into an `engine.pak` file.
-
-The result is a project game release layout in the install directory that's located at `<install>\bin\Windows\release\Default`. In your project game release layout, your `engine.pak` file is located in `Cache\pc`.
-
-
-#### Option 2: Monolithic
+The most common workflow is to release your game only, in which case you should choose the **monolithic** build. This will build only your game/server launcher and any necessary dependencies.
```
willihay, line 160: There's a display bug with `{{< note >}}` shortcodes inside of `{{< tabs >}}`. It's causing the Note callout to outdent and the steps below to both be labelled **1.**. Would you mind addressing this while you're here? As a workaround, I'd suggest simply removing the `{{< note >}}` shortcode around the text below that begins "Use `Visual Studio 16` as the generator". You could start it with "**Note:** " instead. Then just make sure that paragraph is indented.

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -208,65 +216,73 @@ cd C:\o3de-install
...
     cd C:\o3de
-    cmake -B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_VERSION_ENGINE_NAME=o3de-install -DLY_MONOLITHIC_GAME=1 -DLY_PROJECTS=C:\o3de-projects\MyProject
+    cmake -B build/windows -G "Visual Studio 16" -DLY_3RDPARTY_PATH=C:\o3de-packages -DLY_VERSION_ENGINE_NAME=o3de-install -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_PROJECTS=C:\o3de-projects\MyProject
     ```
-    - Include `-D` option, `LY_PROJECTS`, to point to your project's path.
 
-2. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
+    {{< note >}}
+Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
+    {{< /note >}}
```
willihay, line 261: Same `{{< note >}}` shortcode bug as mentioned earlier in this topic. (Only this time I can submit a review comment :) .)
```suggestion
    **Note:** Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
```

willihay (changes requested):

content/docs/user-guide/packaging/windows-release-builds.md
lumberyard-employee-dm: Yes the Server Launcher will be built as well.

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -208,65 +216,73 @@ cd C:\o3de-install
...
+The most common workflow is to release your game only, in which case you should choose the **monolithic** build. This will build only your game/server launcher and any necessary dependencies.
+
+1. Use CMake to create a Visual Studio project for an engine that supports monolithic projects. Specify a new CMake build directory (`build\windows_mono`) in your pre-built SDK engine directory, separate from your non-monolithic build directory. Specify a monolithic build by enabling the `-D` option, `LY_MONOLITHIC_GAME`.
 
     ```cmd
     cd C:\o3de
-    cmake -B build/windows -G "Visual Studio 16" -DLY_3RDPARTY_PATH=C:\o3de-packages -DLY_VERSION_ENGINE_NAME=o3de-install -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_PROJECTS=C:\o3de-projects\MyProject
+    cmake -B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_VERSION_ENGINE_NAME=o3de-install -DLY_MONOLITHIC_GAME=1 -DLY_PROJECTS=C:\o3de-projects\MyProject
     ```
+    - Include `-D` option, `LY_PROJECTS`, to point to your project's path.
```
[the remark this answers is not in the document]
lumberyard-employee-dm: Actually better yet, simpler instructions are to tell the users to use the CMake Presets that O3DE provides for engine builds.
Users can specify the [windows-mono-default](https://github.com/o3de/o3de/blob/development/cmake/Platform/Windows/CMakePresets.json#L20-L28) configuraiton preset which will automatically create the appropriate [build directory](https://github.com/o3de/o3de/blob/development/cmake/Platform/Windows/CMakePresets.json#L62) and set the [LY_MONOLITHIC_GAME to ON](https://github.com/o3de/o3de/blob/development/cmake/Platform/Common/CMakePresets.json#L80-L90)

The instructions can then be simplified to
```
cmake --preset windows-mono-default -DLY_VERSION_ENGINE_NAME=o3de-install
```

The `LY_VERSION_ENGINE_NAME` option is still needed as the preset doesn't set the engine name value that would be used in the SDK layout `engine.json`.

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -208,65 +216,73 @@ cd C:\o3de-install
...
-Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Studio 17` for Visual Studio 2022. For a complete list of common generators for each supported platform, refer to [Configuring projects](/docs/user-guide/build/configure-and-build/#configuring-projects).
-    {{< /note >}}
-
-1. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
+2. In your source engine, use CMake to invoke Visual Studio to append non-monolithic release artifacts to the pre-built SDK layout.
 
     ```cmd
     cd C:\o3de
-    cmake --build build/windows --target INSTALL --config release
+    cmake --build build\windows_mono --target INSTALL --config release
```
lumberyard-employee-dm, line 237: The build step can also be slightly simplified by using the CMake build presets feature.
```suggestion
    cmake --build --preset windows-mono-install  --config release
```

The `--config` option still needs to be specified because we don't configure it with the the windows-mono-install preset.
(Side Note: We can always add a "windows-mono-install-release" preset that inherits from the [windows-mono-install](https://github.com/o3de/o3de/blob/development/cmake/Platform/Windows/CMakePresets.json#L172) build preset and the [release](https://github.com/o3de/o3de/blob/development/cmake/Platform/Common/CMakePresets.json#L192-L198) build preset to make this even simpler for users)

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -208,65 +216,73 @@ cd C:\o3de-install
...
 
-1. Use CMake to create a Visual Studio project for an engine that supports monolithic projects. Specify a new CMake build directory (`build\windows_mono`) in your pre-built SDK engine directory, separate from your non-monolithic build directory. Specify a monolithic build by enabling the `-D` option, `LY_MONOLITHIC_GAME`.
+A pre-built SDK engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for the engine and registered the engine.
+
+1. Reconfigure your source engine with the `-D` option, `LY_PROJECTS`, pointing to your project's path.
 
     ```cmd
     cd C:\o3de
-    cmake -B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_VERSION_ENGINE_NAME=o3de-install -DLY_MONOLITHIC_GAME=1 -DLY_PROJECTS=C:\o3de-projects\MyProject
+    cmake -B build/windows -G "Visual Studio 16" -DLY_3RDPARTY_PATH=C:\o3de-packages -DLY_VERSION_ENGINE_NAME=o3de-install -DCMAKE_INSTALL_PREFIX=C:\o3de-install -DLY_PROJECTS=C:\o3de-projects\MyProject
```
lumberyard-employee-dm, line 256: Again to `LY_PROJECTS` option should not be used when adding a build permutation to the SDK layout.

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -208,65 +216,73 @@ cd C:\o3de-install
...
-#### Option 1: Non-monolithic
+Now, you will choose a build type for your project game release layout.
 
-A pre-built SDK engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for the engine and registered the engine.
+#### Option 1: Monolithic (Game only)
 
-1. Reconfigure your source engine with the `-D` option, `LY_PROJECTS`, pointing to your project's path.
+The most common workflow is to release your game only, in which case you should choose the **monolithic** build. This will build only your game/server launcher and any necessary dependencies.
+
+1. Use CMake to create a Visual Studio project for an engine that supports monolithic projects. Specify a new CMake build directory (`build\windows_mono`) in your pre-built SDK engine directory, separate from your non-monolithic build directory. Specify a monolithic build by enabling the `-D` option, `LY_MONOLITHIC_GAME`.
```
lumberyard-employee-dm, line 225: Actually  there are some critical missing instructions in this section.

For a Pre-Built SDK is it assumed the user already has the monolithic artifacts as part of the SDK?

According to the instruction it assumes the user does not.
The configuration command and build command below would create a windows monolithic permutation.

That would be added to the SDK layout folder.

From there the user can cmake configure their O3DE project with the updated SDK folder in a project-centric workflow to generate a solution that can create the project game release layout.

I think we can have a blurb as follows:
> Now that the monolithic build artifacts have been added to the SDK engine layout, your project can configure against that layout to build mono.
> Now switch to your project's source directory and use CMake to create a Visual Studio project for the project that builds against the monolithic engine
```
cd C:\o3de-projects\MyProject
cmake --B build\windows_mono -S , -DCMAKE_INSTALL_PREFIX=C:\o3de-projects\MyProjectGameLayout`
```

Afterwards invoke the CMake build wrapper to build the project game release layout using the `install` target.
```
cmake ---build build\windows_mono --target install --config release
```

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -194,6 +181,27 @@ Use `Visual Studio 16` as the generator for Visual Studio 2019, and `Visual Stud
...
+
+The other option is to release your game plus the **O3DE Editor** and all of its tools, which is is a **non-monolithic** build. This is not recommended since the size of all these tools combined will be much larger than the **monolithic** build.
+
+A source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
+
+To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `INSTALL` target and `release` configuration.
+
+```cmd
+cd C:\MyProject
+cmake --build build\windows --target INSTALL --config release
```
lumberyard-employee-dm, line 196: Can you replace all instances of `INSTALL` with `install`?

The INSTALL target is only applicable when using an MSVC project, while the `install` target works with MSVC projects, ninja, make and Xcode projects.
This would reduce the user needing to learn to use a different target name for Linux or Mac.

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Good call-out, it's not really a source engine build though. Maybe "Building from a source engine supports non-monolithic projects by default."

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Correct, will update, thanks!

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Good catch! I think that might be too far down to mention it though because they need to run it in this section, so I can add the options in this section. Let me know if you think it doesn't flow well and we can revamp.

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Thanks I didn't notice that! Yep I can do as you suggested here and below!

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Thanks will do the same fix here!

content/docs/user-guide/packaging/windows-release-builds.md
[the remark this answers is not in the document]
cgalvan: Nice, thanks! Updating

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Good call!

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: I think I can change this to use a preset as well:

`cmake --preset windows-default -DLY_VERSION_ENGINE_NAME=o3de-install`

Let me know if that doesn't work.

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Good call, fixed!

content/docs/user-guide/packaging/windows-release-builds.md
lumberyard-employee-dm: The  `windows-default` prefix will work for the non-monolithic flow.

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Thanks, will fix!

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -187,16 +189,16 @@ Your release build is now complete! Next, you need to [bundle your content](#bun
 
 The other option is to release your game plus the **O3DE Editor** and all of its tools, which is is a **non-monolithic** build. This is not recommended since the size of all these tools combined will be much larger than the **monolithic** build.
 
-A source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
+Building from a source engine supports non-monolithic projects by default. As detailed in [Prerequisites](#prerequisites), you should have already created a Visual Studio project for your project and registered your project to your engine.
 
-To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `INSTALL` target and `release` configuration.
+To build your non-monolithic project for release, use CMake to invoke Visual Studio builder. Specify the `install` target and `release` configuration.
```
lumberyard-employee-dm, line 194: Actually the Visual Studio builder is "[MSBuild](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild?view=vs-2022)" so we should use that term instead.

content/docs/user-guide/packaging/windows-release-builds.md
```diff
@@ -222,21 +224,29 @@ Now, you will choose a build type for your project game release layout.
...
     cd C:\o3de
-    cmake --build build\windows_mono --target INSTALL --config release
+    cmake --build --preset windows-mono-install --config release
     ```
 
+1. Now that the monolithic build artifacts have been added to the SDK engine layout, your project can configure against that layout to run a monolithic build. Switch to your project's source directory and use CMake to create a Visual Studio project for your project that builds against the monolithic engine.
+
+    ```cmd
+    cd C:\MyProject
+    cmake --B build\windows_mono -S . -DCMAKE_INSTALL_PREFIX=C:\MyProject\MyProjectGameLayout
```
lumberyard-employee-dm, line 245: Also the `-DLY_MONOLITHIC_GAME=1` will be needed here, as the project needs to be told to use those monolithic artifacts as well.

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Will do!

content/docs/user-guide/packaging/windows-release-builds.md
cgalvan: Good catch, I'll add!

lumberyard-employee-dm (approved):

willihay (approved):

PR #2059, opened 2022-11-02
add test best practices

Kadino:
Signed-off-by: sweeneys <sweeneys@amazon.com>

## Change summary

Added a documentation page for high-level testing advice.

As there may be a lot of feedback on this material, I'm planning a round of discussion from @o3de/sig-testing before inviting docs maintainers for feedback.

### Submission Checklist:

* [ ] **Descriptive active voice** - Do descriptive sentences have a clear *subject* and *action verb*?
* [ ] **Answer the question at hand** - Does the documentation answer a *what*, *why*, *how*, or *where* type of question?
* [ ] **Consistency** - Does the content consistently follow the [Style Guide](https://o3de.org/docs/contributing/to-docs/style-guide/quick-reference)?
* [ ] **Help the user** - Does the documentation show the user something *meaningful*?

PR #2184, opened 2023-01-06
Test best practices revisions from public feedback

Kadino:
Signed-off-by: sweeneys <sweeneys@amazon.com>

## Change summary

Minor grammar, spelling, comment, and structure updates suggested by reviews after sharing in other online software communities.

### Submission Checklist:

### DIFF
--- a/content/docs/user-guide/testing/best-practices/_index.md
+++ b/content/docs/user-guide/testing/best-practices/_index.md
@@ -172,7 +172,7 @@ TEST(MathTests, FastSquare_Integer_Squared1)
 TEST(MathTests, FastSquare_Integer_Squared2)
 {
     // relies on a duplicate implementation of the production code which contains
-    // the same bug, and only helps prove the buggy implementation was not changed
+    // the same bug, and only helps prove the production bug was not changed
     int tenSquared = CalculateTestSquare(10);
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, tenSquared);
@@ -180,8 +180,8 @@ TEST(MathTests, FastSquare_Integer_Squared2)
 
 TEST(MathTests, FastSquare_Integer_Squared3)
 {
-    // uses incorrectly computed squares from aonther file which have the production
-    // bug baked in, and only helps prove the buggy implementation was not changed
+    // uses incorrectly computed squares from another file, and only helps
+    // prove the production bug was not changed
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, pcs_ten);
 }
--- a/content/docs/user-guide/testing/best-practices/_index.md
+++ b/content/docs/user-guide/testing/best-practices/_index.md
@@ -91,18 +91,20 @@ It can be okay to include negative assertions when there is also at least one po
 
 ### Disabling and skipping tests
 
-Ideally, all automated test failures prompt immediately fixing the broken feature (or updating the test when underlying expectations have changed). However, sometimes a tradeoff must be made to break a feature while enabling other functionality. In a case where functionality is known to be broken, automated tests can be temporarily "skipped" or disabled.
+Ideally, all automated test failures prompt immediately fixing the broken feature, or updating the test when underlying expectations have changed. However, sometimes a tradeoff must be made that breaks a feature to enable other functionality. In a case where functionality is known to be broken, automated tests can be temporarily "skipped" or disabled.
 
-Avoid immediately disabling tests when you see a failure. First consider what action the failure blocks, and why all contributors ignoring failure would be less risky. Whenever a test provided with O3DE gets disabled, please [cut an issue](https://github.com/o3de/o3de/issues/new/choose) to track fixing the feature and re-enabling the test.
+Avoid immediately disabling tests when you see a failure. First consider what action the failure blocks, and how critical it is to unblock. Compare that with the risk tradeoff from all contributors ignoring all future bugs the test could catch.
+
+Whenever a test provided with O3DE gets disabled, please [create an issue](https://github.com/o3de/o3de/issues/new/choose) to track fixing the feature and re-enabling the test.
 
 ## Practices to remember
 
 ### Insanity checks
 
-Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way the test should detect, then run the test suite. If no failure occurs, investigate why!
+Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way that the test should detect, and then run the test suite. If no failure occurs, investigate why!
 
 {{< caution >}}
-Immediately revert any intentionally broken code after verifying failure can be detected.
+Immediately revert any intentionally broken code, after verifying failure can be detected by the test.
 {{< /caution >}}
 
 ### Floating-point assertions
@@ -162,8 +164,7 @@ int CalculateTestSquare( const int number)
 
 TEST(MathTests, FastSquare_Integer_Squared1)
 {
-    // despite the name suggesting this verifies the math of computing a square
-    // this test trusts the function is correct, and only proves an identity property
+    // Assumes the function is correct, thus only proves an identity property
     int tenSquared = Math::FastSquare(10);
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, tenSquared);
@@ -171,8 +172,8 @@ TEST(MathTests, FastSquare_Integer_Squared1)
 
 TEST(MathTests, FastSquare_Integer_Squared2)
 {
-    // relies on a duplicate implementation of the production code which contains
-    // the same bug, and only helps prove the production bug was not changed
+    // Relies on a duplicate implementation of the production code that contains
+    // the same bug, thus only proves the production bug was not changed
     int tenSquared = CalculateTestSquare(10);
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, tenSquared);
@@ -180,15 +181,15 @@ TEST(MathTests, FastSquare_Integer_Squared2)
 
 TEST(MathTests, FastSquare_Integer_Squared3)
 {
-    // uses incorrectly computed squares from another file, and only helps
-    // prove the production bug was not changed
+    // Uses incorrectly computed squares from another file, thus only
+    // proves the production bug was not changed
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, pcs_ten);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared4)
 {
-    // directly states that 10^2 = 20, which is incorrect but also easy to notice
+    // directly states that 10^2 = 20, which is incorrect but also *easy to notice*
     int tenSquared = 20;
     int result = FastSquare(10);
     ASSERT_EQ(result, tenSquared);
@@ -196,7 +197,7 @@ TEST(MathTests, FastSquare_Integer_Squared4)
 
 TEST(MathTests, FastSquare_Integer_Squared5)
 {
-    // tersely verifies 10^2 results in 20, which is incorrect but should be easy to notice
+    // tersely verifies 10^2 results in 20, which is incorrect but should be *easy to notice*
     ASSERT_EQ(FastSquare(10), 20);
 }
 ```
@@ -213,7 +214,13 @@ The most straightforward automated tests are unit tests. Unit tests have a very 
 
 The purpose of a unit test is to be small, specific, and fast. They are easy to run, and when they fail, it tends to be easy to determine why. This ease of use primarily comes from the "one call" aspect. Different interactions are treated as additional tests rather than different steps within the same test with a more extended workflow. Think of a unit test as creating an extremely simple robot that quickly reports if one specific behavior breaks. Each robot allows the author to reduce expending mental resources worrying about a design concern. It also documents and monitors that concern for anyone unfamiliar with the area. This allows developers to focus on changing production code while trusting their robots will efficiently help them detect and debug issues as they get introduced. Consider unit tests when writing code intended to exist for more than a week.
 
-A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***. First set up your environmental state (Arrange), then call the function under test (Act), and finally verify the expected side effects occurred (Assert). This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
+A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***:
+
+1. Set up your environmental state (Arrange)
+2. Call the function under test (Act)
+3. Verify the expected side effects occurred (Assert)
+
+This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
 
 ```cpp
 TEST(VectorTests, PushBack_ContainsTwoItems_ThreeTotalItems)
@@ -233,7 +240,7 @@ def test_ListAppend_ContainsTwoItems_ThreeTotalItems(self):
     assert 3 == len(testList)  # Assert
 ```
 
-Note that neither of the preceding tests would be good candidates to include with O3DE, as they test the behavior of non-O3DE libraries (STL containers and Python built-ins). External dependencies should have their own automated tests which live in their own projects. When writing a unit test, focus on one specific behavior of *your* code. Here are real-world examples of O3DE's unit tests in [C++](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Tests/Geometry2DUtils.cpp) and [Python](https://github.com/o3de/o3de/blob/development/Tools/LyTestTools/tests/unit/test_codeowners_hint.py). These tests take milliseconds to verify automatically!
+Note that neither of the preceding tests are good candidates to include with O3DE, as they test the behavior of non-O3DE libraries (STL containers and Python built-ins). External dependencies should have their own automated tests that live in their own projects. When writing a unit test, focus on one specific behavior of *your* code. Here are real-world examples of O3DE's unit tests in [C++](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Tests/Geometry2DUtils.cpp) and [Python](https://github.com/o3de/o3de/blob/development/Tools/LyTestTools/tests/unit/test_codeowners_hint.py). These tests take milliseconds to verify automatically!
 
 ### Integration tests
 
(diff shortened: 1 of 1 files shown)


cgalvan (approved):

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +162,41 @@ int CalculateTestSquare( const int number)
 
 TEST(MathTests, FastSquare_Integer_Squared1)
 {
+    // despite the name suggesting this verifies the math of computing a square
+    // this test trusts the function is correct, and only proves an identity property
```
chanmosq, line 166: I don't think the first line is necessary. Other than that, I have some suggestions on the wording for this and the tests below, which includes omitting the word "only"
```suggestion
    // Assumes the function works correctly and proves an identity property
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +162,41 @@ int CalculateTestSquare( const int number)
...
     int tenSquared = Math::FastSquare(10);
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, tenSquared);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared2)
 {
+    // relies on a duplicate implementation of the production code which contains
+    // the same bug, and only helps prove the production bug was not changed
```
chanmosq, line 175:
```suggestion
    // Relies on a duplicate implementation of the production code that contains
    // the same bug, and proves the production bug was not changed
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +162,41 @@ int CalculateTestSquare( const int number)
...
     int tenSquared = CalculateTestSquare(10);
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, tenSquared);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared3)
 {
+    // uses incorrectly computed squares from another file, and only helps
+    // prove the production bug was not changed
```
chanmosq, line 184:
```suggestion
    // Uses incorrectly computed squares from another file, and 
    // proves the production bug was not changed
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +162,41 @@ int CalculateTestSquare( const int number)
...
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, pcs_ten);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared4)
 {
+    // directly states that 10^2 = 20, which is incorrect but also easy to notice
```
chanmosq, line 191:
```suggestion
    // Verifies that 10^2 = 20, which is an incorrect expectation
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +162,41 @@ int CalculateTestSquare( const int number)
...
     int tenSquared = 20;
     int result = FastSquare(10);
     ASSERT_EQ(result, tenSquared);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared5)
 {
+    // tersely verifies 10^2 results in 20, which is incorrect but should be easy to notice
```
chanmosq, line 199:
```suggestion
    // Concisely verifies if 10^2 = 20, which is an incorrect expectation
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -190,15 +205,15 @@ The incorrect math of the production code is trusted in every test above; all te
...
 
-A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***. First, set up your environmental state (Arrange), then call the function under test (Act), and finally, verify the expected side effects occurred (Assert). This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
+A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***. First set up your environmental state (Arrange), then call the function under test (Act), and finally verify the expected side effects occurred (Assert). This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
```
chanmosq, line 216: - I think having the commas is appropriate here. The problem might be using commas to separate the steps. When listing sequential steps in one sentence, they can be joined together using semicolons instead. 
- The colon after "Arrange, Act, and Assert" simply indicates that the following expands from it.
- Don't need to italicize and bold. I think only italicizing is fitting, since that's what we use to emphasize a term or method.
```suggestion
A good pattern to follow when writing unit tests is *Arrange, Act, Assert*: first, set up your environmental state (Arrange); then, call the function that's under test (Act); finally, verify that the expected side effects occurred (Assert). This simple structure leads to short, focused tests that are easy to understand. Below is a simple example in C++:
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -218,7 +233,7 @@ def test_ListAppend_ContainsTwoItems_ThreeTotalItems(self):
...
+Note that neither of the preceding tests would be good candidates to include with O3DE, as they test the behavior of non-O3DE libraries (STL containers and Python built-ins). External dependencies should have their own automated tests which live in their own projects. When writing a unit test, focus on one specific behavior of *your* code. Here are real-world examples of O3DE's unit tests in [C++](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Tests/Geometry2DUtils.cpp) and [Python](https://github.com/o3de/o3de/blob/development/Tools/LyTestTools/tests/unit/test_codeowners_hint.py). These tests take milliseconds to verify automatically!
```
chanmosq, line 236: Preference to use "are" because "would" is generally an ambiguous word, and "would be" can be written more concisely.
```suggestion
Note that neither of the preceding tests are good candidates to include with O3DE, as they test the behavior of non-O3DE libraries (STL containers and Python built-ins). External dependencies should have their own automated tests that live in their own projects. When writing a unit test, focus on one specific behavior of *your* code. Here are real-world examples of O3DE's unit tests in [C++](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Tests/Geometry2DUtils.cpp) and [Python](https://github.com/o3de/o3de/blob/development/Tools/LyTestTools/tests/unit/test_codeowners_hint.py). These tests take milliseconds to verify automatically!
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -323,28 +338,22 @@ Tests should have unique, obvious names to help document what they perform. One
...
-
-Whenever a test provided with O3DE gets disabled, please [cut an issue](https://github.com/o3de/o3de/issues/new/choose) to track fixing the feature and re-enabling the tests.
-
 ## Test-Driven development (TDD)
 
-*Test-Driven Development* (TDD) is a software writing workflow that prompts engineers to iteratively develop code. The Red-Green-Refactor process can carve up questions such as "what should you build next?" into actionable tasks. This helps avoid getting lost in the ambiguity of software design. TDD also has the extra benefit of leaving behind tests targeting critical requirements. Repeating these three steps can help design new features:
+*Test-Driven Development* (TDD) is a software writing workflow that prompts engineers to iteratively develop code. The Red-Green-Refactor process can carve up questions such as "what should be done next?" and "is this complete?" into actionable tasks. This helps avoid getting lost in the ambiguity of software design. TDD also has the extra benefit of leaving behind tests targeting critical requirements. Repeating these three steps can help design new features:
```
chanmosq, line 356:
```suggestion
*Test-Driven Development* (TDD) is a software writing workflow that prompts engineers to iteratively develop code. The Red-Green-Refactor process can lead to questions such as "what should be done next?" and "is this complete?" into actionable tasks. This helps avoid getting lost in the ambiguity of software design. TDD also has the extra benefit of leaving behind tests that target critical requirements. Repeating these three steps can help design new features:
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -357,7 +366,7 @@ Whenever a test provided with O3DE gets disabled, please [cut an issue](https://
    * Taking a step back, are there any better approaches?
 
 {{< note >}}
-TDD is one small, effective tool to help with software design. It is best employed alongside other design tools, such as the [SOLID design principles](https://en.wikipedia.org/wiki/SOLID). TDD helps focus on lower-level concerns and prompts asking broader questions. However, it also assumes high-level requirements gathering has already provided direction for a task.
+TDD is one small, effective tool to help with software design. It is best employed alongside other design tools, such as the [SOLID design principles](https://en.wikipedia.org/wiki/SOLID). TDD helps focus on lower-level concerns and prompts asking broader questions. However, it also assumes high-level requirements gathering has already provided initial direction for a task.
```
chanmosq, line 369:
```suggestion
TDD is one small, effective tool to help with software design. It is best employed alongside other design tools, such as the [SOLID design principles](https://en.wikipedia.org/wiki/SOLID). TDD helps focus on lower-level concerns and prompts asking broader questions. However, it also assumes that high-level requirements gathering has already provided initial direction for a task.
```

chanmosq (approved):
Looks good! Left some suggestions for verbiage

chanmosq (changes requested):
Requesting responses/changes regarding the suggestions above

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -89,14 +89,20 @@ Negative assertions are different from negative test cases. A negative test case
 
 It can be okay to include negative assertions when there is also at least one positive assertion, but they should not be relied on alone. Focus on identifying the expected output.
 
+### Disabling and skipping tests
+
+Ideally, all automated test failures prompt immediately fixing the broken feature (or updating the test when underlying expectations have changed). However, sometimes a tradeoff must be made to break a feature while enabling other functionality. In a case where functionality is known to be broken, automated tests can be temporarily "skipped" or disabled.
```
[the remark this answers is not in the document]
Kadino: I don't understand the change to "prompt immediately" here.
"prompt immediately fixing" or "prompt you to immediately fix" would be possible changes.

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -89,14 +89,20 @@ Negative assertions are different from negative test cases. A negative test case
...
+
+Avoid immediately disabling tests when you see a failure. First consider what action the failure blocks, and why all contributors ignoring failure would be less risky. Whenever a test provided with O3DE gets disabled, please [cut an issue](https://github.com/o3de/o3de/issues/new/choose) to track fixing the feature and re-enabling the test.
```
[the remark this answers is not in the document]
Kadino: "the failure" makes this less accurate, it would actually be "all future failures"
Why is it appropriate to swap the order of "all contributors"? The original seemed to flow better. Though perhaps "less risky" should come first as risk is the important topic.
Adding remaining recommendations.

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -89,14 +89,20 @@ Negative assertions are different from negative test cases. A negative test case
...
+
 ## Practices to remember
 
 ### Insanity checks
 
-Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way the test should detect. Then, run the test suite. If no failure occurs, investigate why!
+Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way the test should detect, then run the test suite. If no failure occurs, investigate why!
```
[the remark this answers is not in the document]
Kadino: Is there a reason to prefer "a" to "one"? Should is appropriate, as it is not known whether the failure will be detected. Adding "that" and "and"

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -89,14 +89,20 @@ Negative assertions are different from negative test cases. A negative test case
...
 
 {{< caution >}}
-Immediately revert any intentional code changes after verifying failure can be detected.
+Immediately revert any intentionally broken code after verifying failure can be detected.
```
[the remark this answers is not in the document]
Kadino: Adding comma and "...detected by the test"

content/docs/user-guide/testing/best-practices/_index.md
Kadino: Will accept all these changes except the omission of "only"

content/docs/user-guide/testing/best-practices/_index.md
Kadino: "easy to notice" is the point this example is trying to make. "directly states" is the reason it is easy to notice.

content/docs/user-guide/testing/best-practices/_index.md
Kadino: Again "easy to notice" is the important info, and "should" is the appropriate modifier. "terse" is more accurate, as it's actually a bit harder to parse than the previous example.

content/docs/user-guide/testing/best-practices/_index.md
Kadino: I don't see how adding the colon, semicolons, and commas help this sentence. Perhaps it sould become a bulleted list?

Bold and italics was already recommended and accepted in https://github.com/o3de/o3de.org/pull/2059

chanmosq: The change I'm seeing here is to remove the commas, is that right? I think keeping the commas is appropriate. And the subsequent suggestion I make is to break apart the sentence, "First... then... finally...", so it's not one long run-on sentence.

content/docs/user-guide/testing/best-practices/_index.md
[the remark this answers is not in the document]
chanmosq: "prompt you to immediately fix" makes more sense than the original, I suggest fixing it to use that. You can use "you" - that assumes you're talking about the user/developer.

content/docs/user-guide/testing/best-practices/_index.md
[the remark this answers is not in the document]
chanmosq: Changing it to "all future failures" makes the intent much clearer. The suggestion I made was under the assumption that "...ignoring failure..." was referring to the failure that we are considering to disable. 

Consequently, I originally understood this to mean that "_I, the developer, should consider what action the failure blocks, and why ignoring the failure and choosing not to disable it, will be less risky for all_ other _contributors. By other contributors, I mean those who might pull code in which tests have been disabled, which poses a greater risk to them._"
-- This may not be what this text was trying to say, so I think overall we can improve this sentence so it's less ambiguous. 

When you say "all contributors", do you mean "all other contributors" who may encounter these failures. Or do you mean it to refer to the developer who's reading this, suggesting that they consider simply ignoring all future failures. If the latter, I'd suggest removing "contributors" altogether, and opt to use "you". 
As a suggestion, you can say: 
> First consider what action the failure blocks, and why it's less risky if you simply ignore all future failure.

How does this sound?

content/docs/user-guide/testing/best-practices/_index.md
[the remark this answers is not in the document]
chanmosq: It's hard to tell the difference, so one way we can look at this is by completely taking away "that the test can detect" and comparing the two versions: 

- "Intentionally break the code in one way, and then run the test suite"
- "Intentionally break the code in a way, and then run the test suite"

Now there's a greater contrast between the two sentences. The first version implies a singular or specific way to break the code. However, we don't mean to suggest that there's only one way to break the code, nor are we giving an example of one way to break the code. Looking at the second version, this doesn't make much sense - _in what way?_ This is incomplete because it needs a modifier, like "that the test can detect". So what "a" does is it adds generality. It suggests that there are can be one or multiple ways to do this -- whatever way is used, all that matters is that the broken code is one that the test can detect.

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -89,14 +89,22 @@ Negative assertions are different from negative test cases. A negative test case
...
 ## Practices to remember
 
 ### Insanity checks
 
-Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way the test should detect. Then, run the test suite. If no failure occurs, investigate why!
+Ensure a new test is configured to run and report failures by temporarily editing the production code. Intentionally break the code in one way that the test should detect, and then run the test suite. If no failure occurs, investigate why!
 
 {{< caution >}}
-Immediately revert any intentional code changes after verifying failure can be detected.
+Immediately revert any intentionally broken code, after verifying failure can be detected by the test.
```
chanmosq, line 107: Here, "that" is used as the object of the verb "verifying".
```suggestion
Immediately revert any intentionally broken code, after verifying that the failure can be detected by the test.
```

content/docs/user-guide/testing/best-practices/_index.md
chanmosq: Yep, I like these changes

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +164,40 @@ int CalculateTestSquare( const int number)
...
 {
+    // Uses incorrectly computed squares from another file, thus only
+    // proves the production bug was not changed
     int result = Math::FastSquare(10);
     ASSERT_EQ(result, pcs_ten);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared4)
 {
+    // directly states that 10^2 = 20, which is incorrect but also *easy to notice*
```
chanmosq, line 192: Since this is in a code block, italics don't render 
```suggestion
    // directly states that 10^2 = 20, which is incorrect but also easy to notice
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -155,33 +164,40 @@ int CalculateTestSquare( const int number)
...
     int tenSquared = 20;
     int result = FastSquare(10);
     ASSERT_EQ(result, tenSquared);
 }
 
 TEST(MathTests, FastSquare_Integer_Squared5)
 {
+    // tersely verifies 10^2 results in 20, which is incorrect but should be *easy to notice*
```
chanmosq, line 200:
```suggestion
    // tersely verifies 10^2 results in 20, which is incorrect but should be easy to notice
```

content/docs/user-guide/testing/best-practices/_index.md
```diff
@@ -190,15 +206,21 @@ The incorrect math of the production code is trusted in every test above; all te
...
+
+A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***:
 
-A good pattern to follow when writing unit tests is ***Arrange, Act, Assert***. First, set up your environmental state (Arrange), then call the function under test (Act), and finally, verify the expected side effects occurred (Assert). This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
+1. Set up your environmental state (Arrange)
+2. Call the function under test (Act)
+3. Verify the expected side effects occurred (Assert)
+
+This simple structure leads to short, focused tests which are easy to understand. Below is a simple example in C++:
```
chanmosq, line 223: Looks good!

chanmosq (approved):

content/docs/user-guide/testing/best-practices/_index.md
Kadino: Had added the asterisks for emphasis. Keeping unless you disagree with them rendering as asterisks.

content/docs/user-guide/testing/best-practices/_index.md
chanmosq: I mean to say "that"-clause is the whole object. So "that" is a needed conjunction to the clause "the failure can be detected by the test".

Kadino:
> @Kadino Let me know when you're done addressing feedback, so we can get this merged in

Looks good to me, thanks!

PR #2231, opened 2023-02-03
added source code map for project manager. Starting with initial batc…

tkothadev:
…h of classes to test map format

Signed-off-by: TJ Kotha <112996779+tkothadev@users.noreply.github.com>

Created an initial version of the source code map for Project Manager. The goal is to provide a high level map of all code files with brief descriptions of the functionality and logic, including links to github code and other web pages where necessary. I started with only a small sample of code files in this PR, but plan to add more classes to the map once the format is finalized.

### Submission Checklist:

### DIFF
--- a/content/docs/engine-dev/tools/project-manager/_index.md
+++ b/content/docs/engine-dev/tools/project-manager/_index.md
@@ -13,4 +13,4 @@ The development source for the Project Manager can be found here: https://github
 | Name | Description |
 |-|-|
 | [Bootup Sequence](./bootup-sequence) | Learn about how the Project Manager starts up. This will include all necessary steps in the sequence and identify some key classes that affect the rest of the Project Manager's architecture.|
-| [Source Code Map](./source-code-map) | This contains a high level map of all code files in Project Manager, including brief descriptions and links to further information for each class.|
+| [Source Code Map](./source-code-map) | This contains a high level map of some of the most important code files in Project Manager, including brief descriptions and links to further information for each class.|
--- a/content/docs/engine-dev/tools/project-manager/_index.md
+++ b/content/docs/engine-dev/tools/project-manager/_index.md
@@ -13,3 +13,4 @@ The development source for the Project Manager can be found here: https://github
 | Name | Description |
 |-|-|
 | [Bootup Sequence](./bootup-sequence) | Learn about how the Project Manager starts up. This will include all necessary steps in the sequence and identify some key classes that affect the rest of the Project Manager's architecture.|
+| [Source Code Map](./source-code-map) | This contains a high level map of all code files in Project Manager, including brief descriptions and links to further information for each class.|
--- a/content/docs/engine-dev/tools/project-manager/source-code-map.md
+++ b/content/docs/engine-dev/tools/project-manager/source-code-map.md
@@ -25,7 +25,7 @@ This contains brief summaries of various class's functionality and role. If avai
     - [ScreensCtrl](#screensctrl)  [[header](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/ScreensCtrl.h) ] [[cpp file](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/ScreensCtrl.cpp) ]
         - Container structure for all Project Manager Screens and GUI Widgets. Also houses primary transition code for facilitating Project Manager's business logic.
     - ScreenFactory  [[header](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/ScreenFactory.h) ] [[cpp file](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/ScreenFactory.cpp) ]
-        - Helper class for ScreensCtrl that routes `ProjectManagerScreen` enums to appropriate ScreenWidget constructor, invokes that construtor, and returns an instance of a given screen for the Project Manager's business logic.
+        - Helper class for `ScreensCtrl` that routes `ProjectManagerScreen` enums to appropriate `ScreenWidget` constructor, invokes that constructor, and returns an instance of a given screen for the Project Manager's business logic.
 
 
 ### Application
--- a/content/docs/engine-dev/tools/project-manager/source-code-map.md
+++ b/content/docs/engine-dev/tools/project-manager/source-code-map.md
@@ -1,0 +1,110 @@
+---
+linkTitle: Source Code Map
+title: Source Code Map
+description: Summaries of classes and code files in Project Manager
+---
+
+{{< note >}}
+This information is for developers of the **Project Manager** tool. If you're a user working with Project Manager, please refer to the [Project Manager User Guide](/docs/user-guide/project-config/project-manager).
+{{< /note >}}
+
+{{< note >}}
+This contains brief summaries of various class's functionality and role. If available, there are additional links in the table of contents, either with expanded information, or github links.
+{{< /note >}}
+
+## Map
+- <b>Startup Classes</b>
+    - main.cpp [[cpp file](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/main.cpp) ]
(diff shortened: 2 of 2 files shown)


content/docs/engine-dev/tools/project-manager/_index.md
```diff
@@ -13,3 +13,4 @@ The development source for the Project Manager can be found here: https://github
 | Name | Description |
 |-|-|
 | [Bootup Sequence](./bootup-sequence) | Learn about how the Project Manager starts up. This will include all necessary steps in the sequence and identify some key classes that affect the rest of the Project Manager's architecture.|
+| [Source Code Map](./source-code-map) | This contains a high level map of all code files in Project Manager, including brief descriptions and links to further information for each class.|
```
AMZN-alexpete, line 16:
```suggestion
| [Source Code Map](./source-code-map) | This contains a high level map of some of the most important code files in Project Manager, including brief descriptions and links to further information for each class.|
```

AMZN-alexpete (approved):

content/docs/engine-dev/tools/project-manager/source-code-map.md
```diff
@@ -0,0 +1,110 @@
...
+        - Helper class for ScreensCtrl that routes `ProjectManagerScreen` enums to appropriate ScreenWidget constructor, invokes that construtor, and returns an instance of a given screen for the Project Manager's business logic.
+
+
+### Application
+
+Inherits from `AzFramework::Application`.
+
+The functions of the Application class are exclusively run by `main.cpp`.
+
+The two primary functions to keep track of are `Init` and `Run`. [`TearDown`](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/Application.cpp#L260-L272) is invoked at `Application`'s deconstruction.
```
willihay, line 35: I agree that it's super useful to have links to specific lines and blocks of source code, however at the same time I worry about the confusion it could lead to when the source code inevitably changes and the links no longer line up with the intended line or range. I don't have a good solution at the moment, but I wanted to throw this out there. It is possible to create permalinks to specific commits that will always look correct, but then if the code changes, the viewer won't see the new code.

content/docs/engine-dev/tools/project-manager/source-code-map.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+##### [`Application::Init`](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/Application.cpp#L34)
+* Sets various `QApplication` and `QCoreApplication` attributes and parameters
```
willihay, line 39: Nit: We generally use periods at the end of every line of bulleted lists of information like this.

content/docs/engine-dev/tools/project-manager/source-code-map.md
```diff
@@ -0,0 +1,110 @@
...
+* Gets connection to CommandLine, and parses for screen and projectPath parameters
+* Creates ProjectManagerWindow
+
+##### [`Application::Run`](https://github.com/o3de/o3de/blob/development/Code/Tools/ProjectManager/Source/Application.cpp#L274)
+* Sets up Qt GUI stylesheets
+* Sets up Window decoration and MainWindow Geometry
+* Shows the MainWindow
+* Run the QApplication's main loop
+
+[Back to Top](#map)
```
willihay, line 53: I like the use of a "Back to Top" here. Update all of the links like this one to `#overview` if you accept my earlier header change suggestion.
```suggestion
[Back to Top](#overview)
```

content/docs/engine-dev/tools/project-manager/source-code-map.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+
+### ScreensCtrl
+ScreensCtrl is the central widget which stores all GUI information for Project Manager. 
+
+Screens are typically accessed using `m_screenMap`. All screens are stored in `m_screenStack`. The three primary tabs of Project Manager (Project, Gem, and Engine), are stored in `m_tabWidget`. Traversal history through the screens of ProjectManager can be tracked via `m_screenVisitOrder`.
+
+All functions are summarized as follows:
```
willihay, line 79: This is good information, but I'm going to play devil's advocate here and raise a few challenge questions:

1. Shouldn't all of these comments be in the source code itself? The level of commenting in the source in each function header and within the function body appears rather sparse. It would seem more useful to me to have some of these comments in the code itself, where it's more likely to be updated as implementation changes.
2. Is it worth documenting the function implementation in this much detail here? Two challenges this presents:

- As the source changes, the line numbers will occasionally change and invalidate the links (as mentioned elsewhere in this review).
- Any changes to the implementation of a function would require a change to this documentation, and that is going to present a maintenance challenge for the most detailed part of this documentation to remain correct and up to date.

I'm wondering how to maximize the usefulness of this kind of documentation and minimize the maintenance overhead. Would it be better to limit the focus of source code map content to just an overview of each class and function, followed in some cases by a discussion of how it fits into the overall architecture, including how it's used, and when it plays an important role in various scenarios; and then move the detailed function implementation explanations into the source as comments?

willihay (changes requested):

content/docs/engine-dev/tools/project-manager/source-code-map.md
tkothadev: I've thought about this as well. One potential idea is to embrace the idea that the documentation was written at a certain commit in time, and update all links to match up with that particular commit. This would require saying up at the top what commit the documentation is calibrated for.

This can guarantee accuracy, but anytime a commit significantly changes part of the code, we would have to come back here and recalibrate for the latest commit.

Even with this disadvantage, I think it's better than nothing, because at least it can provide some development context for where the code was at in an earlier point, which can still make it easier to navigate the present code.

content/docs/engine-dev/tools/project-manager/source-code-map.md
tkothadev: For the current state of notes, I completely agree with you that these are basically the same value as in-code comments. I may consider going back to the source code and add these comments in later.

On the other hand, in terms of comprehension, I feel having these comments in the same place as code introduces cognitive burden on code reading, which I've experienced firsthand (and usually goes unsaid by developers). The coding guidelines in the o3de repo also states they would prefer no comments if the code can be made self-documenting instead (which is a work in progress right now).

In principle, having an explanation organized in an external place can make it easier to dive into the code directly and see what is going on, because the text density is much less. I may need to improve the text here so it's brief like a summary. If you've got suggestions on what parts can be pruned, that would be great to hear.

That being said, I feel like the current state of notes would be dramatically improved once I introduce diagrams to better describe the data structures involved. However that will take some time, since I need to make sure they are accurate in what they portray. These notes are there as a temporary stage to that point.

tkothadev: also, per an earlier conversation, if the documentation gets mapped to a particular commit and stated at the top, we no longer have to fear the code making the documentation incorrect/stale, since it's effectively fixed in place.

This will mean however it will not be up-to-date as soon as another commit touches the code, which is a definite disadvantage. But I think it helps us in maintaining clarity in the long run, and we can choose to prioritize updating only for major commits which dramatically change the state of the code.

content/docs/engine-dev/tools/project-manager/source-code-map.md
tkothadev: Based on the commit frequency of the project manager, I believe we can choose to prioritize what commits we update the documentation with. We tend to have more work done closer to stabilization and releasing of a new version of O3DE. As such, major changes at that time make sense to update documentation with.

Another potential idea for the documentation to help combat this issue in the long term would be to provide guidelines/advice on how to navigate the github codebase for new contributors. It's basically a reality that we need to compare and consider diffs, and build up a virtual timeline of what changes were made and why. This can help teach newcomers to identify discrepancies already present in the codebase as well.

tkothadev: sounds good, I will update all links to point to these permalinks (which should point to the relevant commit at the time of doc writing)

willihay (approved):

PR #2300, opened 2023-03-07
Creating asset bundles for patches

micronAMZN:
Added a new topic on creating patch asset bundles. Weight and linktitle updates to topics in this section. Removed overview page and added info to index.

### Submission Checklist:


chanmosq (approved):
Looks good!

content/docs/user-guide/packaging/content-patch-package.md
```diff
@@ -0,0 +1,220 @@
...
+* Create a patched release asset bundle that contains all the game assets and updates.
+* Create a patch asset bundle that contains just the game asset updates.
+
+{{< note >}}
+For most updates, only the game assets will change. When upgrading to a new version of O3DE and pushing an update, you should regenerate all of the existing content to ensure it's properly updated, including the engine assets and auxiliary content.
+{{< /note >}}
+
+### Version the asset list
+
+You must version the existing asset list so that it can be compared to the updated asset list. You can do this by simply adding a version identifier to the name of the existing asset list.
```
AMZN-stankowi, line 48: Added a version to the file name isn't required to do the comparison. It's a step I encourage because it helps with file organization, and it helps teams think forward on the use of these files in asset bundling.

By asking someone to put "v1" into the asset list file name when creating their initial bundles, it forces them to think about things like "what does making a v2 mean?"

content/docs/user-guide/packaging/content-patch-package.md
```diff
@@ -0,0 +1,220 @@
...
+
+1. Add a version identifier to the existing `game_pc.assetlist` in your project's `AssetBundling\AssetLists\` directory:
```
AMZN-stankowi, line 50: I would rather these steps be removed from this tutorial, and instead the first tutorial on asset bundling be updated to include a version in the filename of the asset list.

content/docs/user-guide/packaging/content-patch-package.md
```diff
@@ -0,0 +1,220 @@
...
+In the next section, you'll test the new bundles with your project's release build.
+
+## Test the asset bundles
+
+You can test the asset bundles by simulating two user scenarios.
+
+Because your release build currently has the original v1 asset bundle, start by testing the scenario where a user has your project installed and applies the patch asset bundle containing just the updated assets. Copy the patch asset bundle to your release asset directory.
+
+```cmd
+copy C:\MyProject\AssetBundling\Bundles\game_v1_to_v2_patch_pc.pak ^
```
AMZN-stankowi, line 184: It's probably worth talking about pak loading prioritization here, to help users debug. I don't remember the specifics, but there's a chance that you go through these steps and see no visual change, if the original paks end up taking priority.

The easiest way to handle this would be to not put the new pak file in the wildcard loaded directory, and instead in another directory, and add some C++ to the project that loads that new pak file, specifically after GameApplication initialization, where engine.pak and wildcard pak loading is done.

This is also good because, a more mature project will probably want to manage pak mounting timing more directly, to manage load times of the game and to make sure the right content is initialized at the right time.

Engine.pak is mounted in GameApplication via:
`m_archive->OpenPack("@products@", enginePakPath.Native());`
So a similar call (after getting access to the archive system) would be able to load paks from elsewhere.

content/docs/user-guide/packaging/content-patch-package.md
micronAMZN: Changed to the following:

Because you will compare the updated asset list to the original asset list, it's helpful to version the existing asset list. You can do this by simply adding a version identifier to the name of the existing asset list.

content/docs/user-guide/packaging/content-patch-package.md
micronAMZN: I understand, however, I think this suggestion is out of scope for this tutorial.

PR #2401, opened 2023-05-04
Update SDK to use CMake presets

AMZN-alexpete:
## Change summary

Use CMake presets like the release documentation and include note about how to disable test modules.

### Submission Checklist:


cgalvan (approved):

AMZN-alexpete:
Actually, I'm not sure we want to commit this change because I believe our min CMake version is 3.22 which does not support presets?  Or at least I should update the command with a note in case someone doesn't have that version.

AMZN-Olex (approved):

spham-amzn (approved):

lumberyard-employee-dm (approved):
Good with me.
We can add a `windows-install-profile` build preset that inherits from `windows-install` + `profile` preset to reduce the number of options even more.

micronAMZN (approved):

PR #2437, opened 2023-06-20
Fix version dropdown and warnings

AMZN-daimini:
Fixes issue with version handling where the documentation in docs.o3de.org/docs would not be recognized as being for version 23.05, and the warning header would always be displayed regardless. The url at the top always redirects to the current page.

(note that this fix cannot be verified locally or in the deployment version since it does a check on the base url)

Before:
image

After:
The yellow header should not be displayed in docs.o3de.org, and the version dropdown should display 23.05.

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -5,7 +5,7 @@
 {{ $menu   := site.Menus.main }}
 {{ $isDocs := in .RelPermalink "/docs/" }}
 {{ $isPreview := in site.BaseURL "deploy-preview" }}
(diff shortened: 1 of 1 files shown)


aFinchy (approved):

willihay (approved):

willihay:
I think the config.toml will need to be changed, too. I can update that in a new PR.

AMZN-daimini:
> The config.toml will need to be changed, too, to fix the version selector dropdown. I can update that in a new PR.

I think I have fixed it in the previous PR, but please do make a PR if there's something I missed that needs fixing :)

PR #2479, opened 2023-10-04
update prefab override documentation

alexmontAmazon:
## Change summary

Update the prefab override documentation to reflect new variable names and the switch from registry to cvar settings.

### Submission Checklist:


AMZN-alexpete (approved):

cgalvan (approved):
It might be worth combining these into one section, because they aren't super useful independently (e.g. if you enable in the outliner but not in the inspector, then your changes won't actually trigger to be overrides, and if you only enable in inspector but not outliner, then you won't be able to even select entities in the outliner that overrides would apply to).

alexmontAmazon:
> It might be worth combining these into one section, because they aren't super useful independently (e.g. if you enable in the outliner but not in the inspector, then your changes won't actually trigger to be overrides, and if you only enable in inspector but not outliner, then you won't be able to even select entities in the outliner that overrides would apply to).

That's a good point, but this is at least a step forward. I'm going to merge this and then see about more drastic doc changes...

PR #2530, opened 2023-12-15
Update config.toml for 23.10.1

RoddieKieley:
## Change summary

Added 23.10.1 to the dropdown selector for the main branch and thus the main docs.o3de.org site.

### Submission Checklist:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -36,8 +36,9 @@ sourceRepo = "https://github.com/o3de/o3de"
 # Docs version data format: ["display name", "baseURL"]
 # Note: A version that has "www.o3de.org" in its baseURL will get " (latest)" appended to its display name in the version selector.
 versions = [
-  ["23.10", "https://docs.o3de.org"],
+  ["23.10.1", "https://docs.o3de.org"],
   ["development", "https://development--o3deorg.netlify.app"],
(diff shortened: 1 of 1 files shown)


AMZN-daimini (commented):
I'm not sure what the policy is, but should point release docs just roll under the major release? Cause by this logic we should have other past point releases in the dropdown but we only display the major ones.

I think I'm ok with either decision personally, as long as it's agreed to and documented for future releases :)

Ulrick28 (approved):

AMZN-alexpete (approved):
I'm fine with also listing point releases but in the future, I agree we should discuss if doc updates warrant a separate revision or if we should just update the existing major release page.

adamdbrw (approved):

jhanca-robotecai:
> I'm not sure what the policy is, but should point release docs just roll under the major release? Cause by this logic we should have other past point releases in the dropdown but we only display the major ones.
> 
> I think I'm ok with either decision personally, as long as it's agreed to and documented for future releases :)

Having 2310.1 in the list makes sense as this is the most recent version of the documentation. On the other hand point releases do not break the backward compatibility and there is no reason to use `2310.0` if `2310.1` is available. Maybe `2310` should point to `2310.1` by default?

PR #2541, opened 2024-02-07
Update requirements.md

Matthias-Fauconneau:
CMake does not check for presence of libcurl for AWSCore so this fails later during the build, in any case seems to be a hard dependency. https://github.com/o3de/o3de/issues/11783

### DIFF
--- a/content/docs/welcome-guide/requirements.md
+++ b/content/docs/welcome-guide/requirements.md
@@ -227,11 +227,12 @@ O3DE also requires some additional library packages to be installed:
 * libstdc++-12-dev
 * libunwind-dev
 * libzstd-dev
+* libcurl4-openssl-dev
 
 You can download and install these packages through `apt`.
 
 ```shell
-sudo apt install libglu1-mesa-dev libxcb-xinerama0 libxcb-xinput0 libxcb-xinput-dev libxcb-xfixes0-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev libfontconfig1-dev libpcre2-16-0 zlib1g-dev mesa-common-dev libunwind-dev libzstd-dev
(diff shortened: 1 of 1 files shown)


spham-amzn:
libcurl should not be part of the prerequisites. `libcurl.a` is included in the [3rd party package](https://github.com/o3de/3p-package-source/blob/30e7001c515ed5fe0a596a89e59907c8e7911482/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux#L97C31-L97C40) for AWS Native SDK. This looks like it couldve been a potential error with the 3rd Party [FindAWSNativeSDK.cmake](https://github.com/o3de/3p-package-source/blob/main/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux), however, this error hasn't been repro'd lately.

spham-amzn:
The issue this PR is trying to solve should be fixed in the AWS Native SDK gem itself, by way of doing a [FindCURL](https://cmake.org/cmake/help/latest/module/FindCURL.html) to see if its installed an report the error and its remediation.

ShaunaGordon:
@spham-amzn @Matthias-Fauconneau Is this still relevant? Or has the AWS Native SDK been fixed?

spham-amzn:
> @spham-amzn @Matthias-Fauconneau Is this still relevant? Or has the AWS Native SDK been fixed?

I am fine adding this as a requirement, since it won't break ubuntu. However, the merge needs to be retargeted. It's currently targeting `main`

ShaunaGordon:
As stated in #2589, we won't be including this one, because as it is, it's not a requirement for the officially-supported platforms, and the underlying problem is a bug in the engine, itself and better fixed there.

Thanks for this work, though! It helped us uncover that bug.

PR #2589, opened 2024-08-27
Cherry-pick change from #2541

ShaunaGordon:
CMake does not check for presence of libcurl for AWSCore so this fails later during the build, in any case seems to be a hard dependency. https://github.com/o3de/o3de/issues/11783

## Change summary

#2541 was approved, but required a retarget. The original contributor has been unresponsive lately, so I went ahead to try to retarget it. However, due to changes since his submission, it doesn't rebase cleanly, so I've cherry-picked the change into a fresh copy of development.

From the original PR:

> CMake does not check for presence of libcurl for AWSCore so this fails later during the build, in any case seems to be a hard dependency. https://github.com/o3de/o3de/issues/11783

@spham-amzn can you do a quick check on this, since you were involved with the original PR, please? ❤️ 

### Submission Checklist:

### DIFF
--- a/content/docs/welcome-guide/requirements.md
+++ b/content/docs/welcome-guide/requirements.md
@@ -222,11 +222,12 @@ O3DE also requires some additional library packages to be installed:
 * libunwind-dev
 * libzstd-dev
 * tix
+* libcurl4-openssl-dev
 
 You can download and install these packages through `apt`.
 
 ```shell
-sudo apt install libglu1-mesa-dev libxcb-xinerama0 libxcb-xinput0 libxcb-xinput-dev libxcb-xfixes0-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev libfontconfig1-dev libpcre2-16-0 zlib1g-dev mesa-common-dev libunwind-dev libzstd-dev tix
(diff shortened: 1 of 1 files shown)


ShaunaGordon:
@jhanca-robotecai Done. Are we going to cherry-pick back to dev, then?

spham-amzn:
This is more of a work-around, as I mentioned in https://github.com/o3de/o3de/issues/11783 that this shouldn't be a requirement for the system, and the AWSCore 3P package should either fix this or add detection for it. As a general rule, I'd be reluctant to add a system requirement for a single (optional) gem in the documentation, but I will check the root cause of this dependency issue and if its required by more than AWSCore, I'll be fine with this update. If its only AWS, then I'll look into updating the AWSCore gem.

A similar scenario is the ROS2 gem,  which requires ros packages, but the detection is done only if the gem is activated.

spham-amzn:
So this package is not required by ubuntu, which is the `official` supported platform. This just happens to fix non-ubuntu distributions, so I don't recommend adding this. I also figured out that the bug is in the AWS Native SDK package's FindAWSNativeSDK.cmake, this [line](https://github.com/o3de/3p-package-source/blob/d727740e71bd3f12873853384f0d31c971bda10b/package-system/AWSNativeSDK/FindAWSNativeSDK.cmake.Linux#L115) specifically. The problem is that this should link against the package libcurl.a, not the first libcurl that the system finds.

ShaunaGordon:
Alright, I'll leave it to be fixed as desired in the engine. 👍🏼

PR #2611, opened 2024-10-07
Add generated API docs for 24.09

ShaunaGordon:
## Change summary

Generated API docs, per https://github.com/o3de/sig-docs-community/blob/main/tools/o3de-doxygen/README.md


RoddieKieley (changes requested):
@ShaunaGordon To close the loop here, as per the conversation in Discord it appears there is some issue relating to the boilerplate section listings having 'o3de-doxygen-insert-table' in the newly generated files vs. the actual content in at the very least the development branch online.

Screenshot from 2024-10-08 09-20-02

ShaunaGordon:
Okay, I think I found what the problem was.

Apparently, there's a known issue with `sed` on Mac (and maybe BSD/Unix more broadly) where the `-i` switch is different than on Linux. Sed was failing when I tried running the scripts straight on my Mac. It had looked to be something at the end of the generation, so it looked like it was a non-impacting thing, but it turns out that it did all the generation, but the conversion of things like that table of contents directive weren't done.

Running the scripts in a Docker container appears to have solved it, and the table of contents display as expected now.

RoddieKieley (approved):
Looks good to me now in the spot checking I did locally from the incoming branch :)

PR #2614, opened 2024-10-09
From main into development

ShaunaGordon:
## Change summary

Bring development into alignment with main, now that 24.09 is released.

I assume we'll want to rebase it, or perhaps use a different strategy given the de-syncing that happened until recently, but I figured this was an easy way to see if we would run into any snags.


ShaunaGordon:
If I'm looking at this right, a rebase should be straightforward.

ShaunaGordon:
🥳 I just tested this locally and rebases without issue! 🥳

jhanca-robotecai:
Since we keep the autogenerated files on the repo (this should be changed as soon as possible), we should try to ignore these files as much as possible. This PR is `+1,432,868 −1,226,787`, which makes it impossible to review. 

I would suggest keeping the real changes separated from the autogenerated with the following procedure: 
1. A cherry-pick of the commits that matter as in https://github.com/o3de/o3de.org/pull/2621 
2. A separate PR with autogenerated files only. The `development` codebase should be used when autogenerating the files (cherry-picking the autogenerated files from the `main` will lead to problems when `development` has some code that does not exist in `main`, which could often be the case).

PR #2654, opened 2025-04-28
Correct tiny grammar issues

ShalokShalom:
Correct tiny grammar issues


ShalokShalom:
I applied the rules of Grammarly.

This could serve as a simple standard 🤷🏻‍♂️

JT-SCB-GameDesign:
We are aware of the button issue and have temporarily removed it. Thanks for the heads up.
Issues are better to post plese than PRs for this kind of issue as explained in Discord chat.  So please repost this as an issue and we will log it so it can be evaluated there.

PR #2681, opened 2025-06-07
Hugo Update Step 2 - Extract Theme To Module

ShaunaGordon:
## Change summary

As part of the efforts to clean up the Hugo site and make it easier and less daunting for contributors, I've extracted the theme and made it into a module at https://github.com/o3de/hugo-odie . The remaining theme-related files are specific to the O3DE site and contain site-specific content. Ideally, they'll get pulled out eventually, too, and made into proper templates with site-specific content put into content files. That's a future phase.

In the latest version of the theme, I also took the liberty of commenting out the "edit on Github" button, due to the issues it's causing.

## Testing

To test, use the Netlify preview site, or download the branch and run locally. In either case, compare to `development` and ensure the user experience is the same. The only thing that should be different is this version should not have the "edit on Github" button.


JT-SCB-GameDesign (approved):
Looking good, did the Netlify review.

PR #2683, opened 2025-06-17
Update reference to api.o3de.org repo to v25.05.0

ShaunaGordon:
## Change summary

With the API docs in their own repo, we need to update the reference to it. This PR updates the reference to the newly-generated v25.05.0 api docs.


JT-SCB-GameDesign (approved):
kk

PR #2700, opened 2025-07-22
Release-notes/25.05.1

ShaunaGordon:
## Change summary

Add release notes for 25.05.1 and update references to 25.05 stuff.


jhanca-robotecai (changes requested):
This PR targets the `development` branch, while it looks like a point release (hence it should be targeting `main`)

jhanca-robotecai (approved):
Okay after offline explanations.

Let's cherry-pick it to all branches.

As for the review: it would be an added value to keep the releases in an order, which is not the case for now 
image

Issue #2705, opened 2025-08-10
[DOCS] Lyshine, UI components, dev-languages


mythrz:
## Describe the issue

UI Lyshine and UI components are missing relevant information on how to use the Gem and its components. There are some code examples in the LyShineExamples but these are NOT focused or individualised on the tasks and necessities of the professional. They are all abstract and over-generalised. Ideally, each UI-component page should have clear instruction on how to achieve the expected result

## Some of Impacted pages found so far

https://www.docs.o3de.org/docs/user-guide/interactivity/user-interface/components/working/

https://www.docs.o3de.org/docs/user-guide/programming/components/create-component/

All of the interactive componenents pages (child pages. e.g., button, textinput, dropdown, etc):

https://www.docs.o3de.org/docs/user-guide/interactivity/user-interface/components/interactive/

Some UI Canvases. Accessing UI elements at runtime:

https://www.docs.o3de.org/docs/user-guide/interactivity/user-interface/canvases/accessing-ui-canvas-runtime/

## Expected experience

Clear and concise examples and descriptions on how to use LyShine and its components. 
 
## Issues worth tracking:

- Dropdown: https://github.com/o3de/o3de.org/pull/2706
- Button: https://github.com/o3de/o3de.org/pull/2703
- Dynamic Layout Content (applied to a dropdown): https://github.com/o3de/o3de.org/pull/2707
- Text Input (only one method): https://github.com/o3de/o3de.org/pull/2714
- UiInitializationBus (important function that triggers after all UI is loaded and ready): https://github.com/o3de/o3de.org/pull/2717
- UIElementBus: Which page should this go? https://github.com/o3de/o3de.org/issues/2705#issuecomment-3383037759
- UITextBus: Which page should this go? https://github.com/o3de/o3de.org/issues/2705#issuecomment-3383001839
- Scrollbar:

mythrz:
Not sure where to place the UITextBus. Leaving it here for now, will structure it later!

---

Two of the most common examples of UITextBus is to set it to a new value or get the value from the UI-Text. 

### Lua

```Lua
local currentSelection = UiTextBus.Event.GetText("NameOfElement");

UiTextBus.Event.SetText(m_idOfTextElement, "newValue"); 
```

[...]

PR #2707, opened 2025-08-10
[Docs] Dynamic Layout content populated with cpp

mythrz:
- Dynamic Layout content populated with cpp
- the example expects another ui component like the dropdown to already be present.

## Change summary

Part of https://github.com/o3de/o3de.org/issues/2705

- Covers how to in how to include and inherit the bus
- how to override the functions/methods/events so it compiles without errors
- how to clone/copy the existing option so it matches the new content size

### Submission Checklist:

* [x ] **Descriptive active voice** - Do descriptive sentences have a clear *subject* and *action verb*?
* [ x] **Answer the question at hand** - Does the documentation answer a *what*, *why*, *how*, or *where* type of question?
* [ x] **Consistency** - Does the content consistently follow the [Style Guide](https://o3de.org/docs/contributing/to-docs/style-guide/quick-reference)?
* [ x] **Help the user** - Does the documentation show the user something *meaningful*?


content/docs/user-guide/interactivity/user-interface/components/layout/dynamic-layout.md
```diff
@@ -32,3 +32,103 @@ For a [**LayoutGrid**](components-layout-grid) element, the cell size of the **L
 1. Create a child entity that has an **Image** component. To do this, right-click your layout component in the **Hierarchy** pane and choose **New**, **Element from Prefab**, **Image**.
 
    This image serves as the prototype element that will be cloned and filled with dynamic content.
+
+---
+
+**Cpp**
+
+To create your own UI components, check the [Working with UI Components](https://www.docs.o3de.org/docs/user-guide/interactivity/user-interface/components/working/) page.
```
ShaunaGordon: Internal links should be relative, not absolute.

ShaunaGordon (changes requested):

content/docs/user-guide/interactivity/user-interface/components/layout/dynamic-layout.md
```diff
@@ -32,3 +32,103 @@ For a [**LayoutGrid**](components-layout-grid) element, the cell size of the **L
...
+    /// rest of the code...
+```
+
+In other words, the above code clones or copies the only Option available ("Option1") and populates the remaining childs with the clones giving this as the result. 
+
+![alt text](image.png)
+
+If you use the vector of strings that contains 3 indexes it will modify the text-element of each option and populate the dropdown as expected.
+
+<img width="236" height="157" alt="image" src="https://github.com/user-attachments/assets/07a1ac9b-dd90-4e17-b6d7-8754f114e30a" />
```
ShaunaGordon: Is there a particular reason this is using github's attachment cdn? It would be better if the image is stored in the project and referenced interally, unless there's a compelling reason to do otherwise.

ShaunaGordon (changes requested):

content/docs/user-guide/interactivity/user-interface/components/layout/dynamic-layout.md
mythrz: Probably copy-past it instead of adding the file. No particular reason. Tired I guess (:
Created a new PR to add the file: https://github.com/o3de/o3de.org/pull/2731

Changing to this then. sec
alt text

ShaunaGordon (approved):

PR #2721, opened 2025-09-26
Add math flag to toml, update theme for current one with the rest of the math support

ShaunaGordon:
## Change summary

To activate math syntax rendering, we needed to add `params.math` to the toml file, and update the theme to load MathJax (the default recommendation for Hugo). This does those things (the theme's already been updated).

## To Test

1. Pick a page or make a new one
2. Write some math as [illustrated here](https://gohugo.io/content-management/mathematics/#step-5)
3. Run `hugo server` and view your page
4. The page should display the math in an evidently different way, and right-clicking on it should bring up a custom MathJax context menu

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,5 @@ go 1.24
 
 require (
 	github.com/o3de/api.o3de.org v0.0.0-20250619195622-67020d64f7b5 // indirect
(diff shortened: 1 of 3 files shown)


JT-SCB-GameDesign (approved):
KK, gunna button push approve by just looking it over and w/o testing first so the work can continue.
I'll test in a few days to make sure it's working on a different machine and reply back ASAP if I have any issues.

PR #2726, opened 2025-10-09
Update minimum requirements: linux

jhanca-robotecai:
## Change summary

This PR updates the minimum requirements on Linux platform in the docs, to match the code:
- clang-12 was changed to clang-14
- cmake was updated from 3.22 to 3.24

Additionally, the instructions for installing cmake 3.24 on Ubuntu 22.04 were added and information about Ubuntu 24.04 being in experimental stage was removed. Note: probably we should set Ubuntu 24.04 as default for  the next release.

This PR targets the stabilization branch directly and should be cherry-picked to the development branch after approved.

### Submission Checklist:


content/docs/welcome-guide/requirements.md
```diff
@@ -183,7 +185,7 @@ Install Clang and the GNU C++ Library using the instructions for the version of
 You can install the default version of Clang for Ubuntu 22.04 LTS, which is clang-14. You will also need to install the corresponding [GNU C++ Library](https://gcc.gnu.org/onlinedocs/libstdc++/).
 
 ```shell
-sudo apt install libstdc++-12-dev clang
+sudo apt install -y libstdc++-12-dev clang clang++
```
nick-l-o3de, line 188: I think we're on minimum clang14, which I assum means libstdc++=14-dev too (Not certain).

Not even sure this is correct anyway, since this would install LATEST `clang++` with libstdc12.
Any linux / debian / ubuntu experts know what we should put here?

jhanca-robotecai: > I think we're on minimum clang14, which I assum means libstdc++=14-dev too (Not certain).

`libstdc++` is a part of GCC and has nothing to do with Clang. It is a default C++ toolchain on Ubuntu (linker, assembler, standard library headers, standard library runtime, runtime libraries, etc.). GCC is still default for many Linux distros, including Ubuntu. Even if you install newer GCC versions, Ubuntu sticks to one “system default” `libstdc++` version for the distro lifecycle, because changing the default `libstdc++` would break the linkage (ABI). Basically, switching the toolchain means you need to rebuilt all C++ packages in the distribution. O3DE does not build on neither Ubuntu 22.04 nor 24.04 with `libstdc++13` or newer, it would require rebuilding all dependencies (e.g. `libxcb`) with the same C++ toolchain.

You might install a C++ toolchain developed by LLVM on Ubuntu if you wish, it is called `libc++`, but it is opt-in (even after installing you still need to force the compiler to use it: `clang++ -stdlib=libc++ main.cpp -o main`). I haven't tried that with O3DE.

Ubuntu 22.04 was the first Ubuntu with GCC 12. Ubuntu 24.04 was released before GCC 13 became mature (LTS releases are quite conservative). Both Ubuntu 22.04 and Ubuntu 24.04 use `libstdc++-12`. This instruction describes installation steps for 22.04 and 24.04, hence it is correct.

> Not even sure this is correct anyway, since this would install LATEST clang++ with libstdc12.

Not latest, but default. This means `clang-14` on 22.04 and `clang-18` on 24.04. But, as explained earlier, they both use the same toolchain.

jhanca-robotecai:
Rebased on top of the stabilization branch.

nick-l-o3de (approved):

Issue #1490, opened 2022-03-17
[ISSUE][DOCS] Entity Inspector documentation page contains deprecated image and information


LB-ArturZieba:
## Describe the issue

There is a deprecated image of two components on the Entity Inspector documentation page:
![component-entity-inspector](https://user-images.githubusercontent.com/86952082/158786939-fdec9cea-6833-452a-9a7c-257cb5f72fb7.png)
Current Entity Outliner:
![EntityOutliner](https://user-images.githubusercontent.com/86952082/158787025-3f28cd6a-1fdd-43e9-ab49-68dee836094d.png)
Current Entity Inspector:
![EntityInspector](https://user-images.githubusercontent.com/86952082/158787069-77c58053-9a01-4c5f-9d40-92f5250de55b.png)

There is also deprecated information under the "To open the Entity Inspector" section - `Status - Active status of the entity. When the level starts, the entity can be active, inactive, or active but only in editor mode.`. Currently the available statuses of entities are "Universal" and "Editor only".

## Impacted page(s)

https://www.o3de.org/docs/user-guide/editor/entity-inspector/

## Expected experience

Documentation contains current images and information.

GaianHelmers:
This can also use the inclusion of the new Status functionality of "Universal" and "Editor Only".
How it causes the EditorOnly entities to not be created at runtime. Most effectively used to remove a redundant prefab wrapper at spawntime.

I cover the use of Editor Only in my O3DE Intensive - Prefabs and Spawning, Timestamped: https://youtu.be/qWS3SPQR5G4?si=xlGJ_MckHEP1DkHR&t=1151

GaianHelmers:
I will update this with my PR: https://github.com/o3de/o3de.org/pull/2745

PR #2745, opened 2025-11-11
Entity Activation System Docs

GaianHelmers:
Docs companion for github.com/o3de/o3de/pull/19319/

Covering
- Basic Entity Activation, and how to handle it directly in code.
- What the entity does to determine activation.
- How the local, and parent activation types affect its state.
- How to register a new active type.
- The future possibilities for the system. How to think about expanding it for your needs.

And, describing that entity activation can happen all the time, and that entities will affect their children in the updated pages.

_Note: This is a draft PR for feedback on the info before refinement._

## Change summary

- Updated inspector, and components overview pages
- Introduced dedicated "Entity Activation System" page for the end of /Components/ section.

### Submission Checklist:


content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,90 @@
+---
+linkTitle: Entity Activation System
+title: Understanding Entity Activation and the Activation Type System
+description: Learn how to set Entity Active States, and define new Custom Activation Types.
+weight: 900
+---
+
+An *entity* in **Open 3D Engine (O3DE)** is the most primordial element in the entire system. As O3DE is an [*entity-component system (ECS)* engine](overview), the *entity* is the core of how everything works.
```
nick-l-o3de: I would recommend not calling it an ECS - that actually has a very specific meaning, which O3DE is not.  

O3DE has entities, and they have components.

But an ECS is a very different beast, and usually involves isolating all of the computation part of the components to **systems** which iterate over bulk arrays of component data using[ Data Oriented Design](https://en.wikipedia.org/wiki/Data-oriented_design) principles.

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,90 @@
...
+
+An Entity has many purposes, but one is that it is the authority of whether it is active or not. As it processes changing its active state, it sets its components accordingly, reports its change of state, and handles bringing itself to a functional place in its new state. Outside systems rely on its state handling, and only instruct it to change, allowing it to handle it all itself.
```
nick-l-o3de: IMO we should be specific here, "it has many purposes" is kinda vague.  I'd prefer specifying exactly what it does

An Entity is made up of a list of Components and a state, which can be (Constructed, Initialized, Active, Inactive).  When the Entity's state changes state, the Entity calls the appropriate functions on the components in its list (Initialized(), Activated(), Deactivated() respectively) to let them know about the state transition.

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Activating and Deactivating an Entity
+
+### Scripting and GameEntityContext
+There are easy to use Script Canvas Nodes to handle Entity Activation: ActivateGameEntity and DeactivateGameEntity nodes.
```
nick-l-o3de: As per the doc guidelines we don't try to infer that something is easy or hard, we just give the information, as dry as possible.

"The following Script Canvas Nodes can be used to handle entity state":

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,90 @@
...
+
+(NODES)
+
+Using C++, the easiest way to set an entitys active state is to use the Game Entity Context Component entity handling events.
```
nick-l-o3de: Similar comment as above

In C++, `AzFramework::GameEntityContextRequestBus` can be used to set an Entity active state:
  * AzFramework::GameEntityContextRequestBus::Events::ActivateGameEntity
  * AzFramework::GameEntityContextRequestBus::Events::DeactivateGameEntity

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: I can remove it. Yeah I know ECS is loaded, I was just quoting the Overview which says that _italic_ piece verbatim.

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: Added! 👌

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: Awesome thanks, got it in!

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: Awesome thanks, got this in too!

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,88 @@
...
+The purpose of this segmentation is to allow systems to do many changes to an Entitys state, then once finalized apply it once. Removing any excessive update calls.
+
+## How an Entity determines it's Active State
+An entity, instead of having one setting of "Active/Inactive" now has a list of up to 32 active state flags.
+By evaluating these many possible flags, an entity can determine if it's active: All flags are true (active), (111111...).
+And determine if it is inactive: All or some flags are false (inactive), (110101...)
+
+By using this "many-state" method of evaluating an Entitys active state, we are able to stack many outside forces that may want control over when an entity is active or inactive without overwriting the entities state before this potential system affected the Entity.
+
+This purpose can be seen in the following examples.
```
nick-l-o3de: I'd suggest just drying it out and relating how it works now, and save the above section for later, (the expert stuff at the bottom, the common stuff at the top), so that the basic user can read as far as they need to in order to get started.

Since its a ussers guide we first always focus on what a normal first-time user might want to know ASAP.

Maybe something like:

### Activating and Deactivating Entities

(just show the API for activating and deactivating an entity from the users perspecive, ie, whatever you would invoke in order to change channel 0)

Note that when deactivating an entity that has transform children, their children, will also automatically deactivate.  Reactivating the entity will restore their children to whatever activation state they were at when the parent reactivates.

### Internal working and future expansion

Entity activation state is controlled indirectly.  At runtime, the Entity contains a 32-bit set of user-defined flags, any of which can be registered by other systems to use.  After changing the bits that are set, `Entity::ApplyEffectiveActiveState` is invoked to compute the state transition.   The entity will transition to Active if all bits are set (which is the default) or Inactive if any bit is unset.   Systems that want to manage entity activation state can register a bit index to control, and then toggle those bits to activate and deactivate entities.  

(document API calls for registering a bit)
(document API call for setting/unsetting a bit in the channels)

As an example, one bit is already reserved for whether the entity itself should be active or not (explicitly), and another bit is reserved by the Transform system (which stores the parent/child relationships between entities).   The transform system controls one of the bits in the bitset related to whether a parent is active or not.  

(explanation of how the transform system uses its bit)

GaianHelmers: Awesome, I think I captured what you're gathering here. I think it makes a lot of sense.

GaianHelmers:
I wonder if I should bump the page up one tier and have it on the root "Programming" section... as it mostly has nothing to do with components. And at the end of the components seems even less relevant. Talking about reflection and behaviour contexts and then jumping into entity functionality.

I find it a little strange that there is no definitive place for "Entities" data and information.
EAS covers basic purpose and then the activation, so it can sorta be an all in one. But again, seems to then make sense as a Programming base entry.

There's also no formal document around how the spawning of entities works.
Which could include basic spawning
How the tickets work and what you can do with them
The actual stage by stage spawning process. Loading the data, cloning, init, and activation.

Then working out async spawning would be able to add to the basic spawning, and spawn process pages when implemented.

Also Prefabs only base is in "Interactivity" which I don't think is apt, and is blank. Could also be part of Entities.

Something like:
```
- Programming
    - Gems
    - Entities
        - Overview
        - Prefabs
        - Spawning Entities
        - Spawn Tickets
        - Entity Spawning System
        - Entity Activation System
    - Components
```

I gathered this topic of discussion into an ideas post as I feel this is a different task than the simple introduction of docs covering this feature addition.
https://github.com/o3de/o3de.org/discussions/2746

GaianHelmers:
Related issue: Entity Inspector dated Images: https://github.com/o3de/o3de.org/issues/1490
I can replace, once we get the interface set how we want it.

GaianHelmers:
I ignored updating the images, but the Entity Activation System https://github.com/o3de/o3de/pull/19319 now fully replicates the functionality described on the **Entity Inspector** page.

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,69 @@
+---
+linkTitle: Entity Activation System
+title: Understanding Entity Activation and the Activation Type System
+description: Learn how to set Entity Active States, and define new Custom Activation Types.
+weight: 900
+---
+
+An Entity in **Open 3D Engine (O3DE)** is made up of a list of Components and a state, which can be (Constructed, Initialized, Active, Inactive). When the Entity's state changes, the Entity calls the appropriate functions on the components in its list (Initialized(), Activated(), Deactivated() respectively) to let them know about the state transition.
```
ShaunaGordon:
```suggestion
An Entity in **Open 3D Engine (O3DE)** is made up of a list of Components and a state, which can be `Constructed`, `Initialized`, `Active`, `Inactive`. When the Entity's state changes, the Entity calls the appropriate functions on the components in its list (`Initialized`, `Activated`, `Deactivated` respectively) to let them know about the state transition.
```

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Activating and Deactivating an Entity
+### Scripting and GameEntityContext
+The following Script Canvas Nodes can be used to handle entity state: ActivateGameEntity and DeactivateGameEntity nodes.
```
ShaunaGordon:
```suggestion
The `ActivateGameEntity` and `DeactivateGameEntity` Script Canvas nodes can be used to handle entity state.
```

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,69 @@
...
+
+![Script Canvas ActivateGameEntity and DeactivateGameEntity nodes.](/images/user-guide/programming/components/activate-game-entity-nodes.png)
+
+In C++, `AzFramework::GameEntityContextRequestBus` can be used to set an entity's active state:
+- `AzFramework::GameEntityContextRequestBus::Events::ActivateGameEntity`
+- `AzFramework::GameEntityContextRequestBus::Events::DeactivateGameEntity`
+
+These, like the Script Canvas nodes, handle everything for you directly.
+
+Note that when deactivating an entity that has transform children, their children, will also automatically deactivate. Reactivating the entity will restore their children to whatever activation state they were at when the parent reactivates.
```
ShaunaGordon:
```suggestion
{{<note>}}
When deactivating an entity that has transform children, their children will also automatically deactivate. Reactivating the entity will restore their children to whatever activation state they were at when the parent reactivates.
{{</note>}}
```

Also, this sentence

> Reactivating the entity will restore their children to whatever activation state they were at when the parent reactivates.

feels a bit unclear and could use reworded to make it more clear that the system restores the children to the state they had been in at the time of parent deactivation.

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,69 @@
...
+
+Handling entity activation manually requires two steps to properly work:
+- You need to change its desired state by using the `SetEntityActive`, or `SetEffectiveActiveLayerByTypeIndex` methods.
+- You then need to call its `ApplyEffectiveActiveState` method to make it actually process its new Active State and change to it accordingly.
+
+The purpose of this segmentation is to allow systems to make many changes to an entity's state, then once finalized, apply it once. Removing any excessive update calls.
+
+## How an Entity determines it's Active State
+The entity activation state is controlled indirectly. At runtime, the Entity contains a 32-bit set of user-defined flags, all active by default, any of which can be registered by other systems to use. After changing the bits that are set, `Entity::ApplyEffectiveActiveState` is invoked to compute the state transition. The entity will transition to Active if all bits are set or Inactive if any bit is unset. Systems that want to manage entity activation state can register a bit index to control, and then toggle those bits to activate and deactivate entities.
```
ShaunaGordon, line 37: > The entity will transition to Active if all bits are set or Inactive if any bit is unset.

This sentence is kind of muddying my understanding of how the system works. It kind of makes it sound like my entity will be rendered Inactive if *any* bit is unset, giving the impression that it's still an all or nothing thing.

I'm guessing the intent is that each entity gets assigned a bit within this register that, for that entity is its on/off switch (a good analogy might be a panel of light switches -- each one controls a different light), and so an entity would activate or deactivate based on its *corresponding* bit in the register?

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,69 @@
...
+//! Evaluates the active state flags to set the entity's Activated/Deactivated state.
+bool ApplyEffectiveActiveState();
+
+//! Returns the current effective active state of the Entity.
+bool IsEffectivelyActive();
+```
+
+As an example, one bit is already reserved for whether the entity itself should be active or not (explicitly), and another bit is reserved by the [Transform system](/docs/user-guide/components/reference/transform/) (which stores the parent/child relationships between entities). The transform system controls one of the bits in the bitset related to whether a parent is active or not.
+
+By using `Entity::SetEffectiveActiveLayerByTypeIndex` you toggle the entity's bit of your registered type's index. Using the example above you would have 2 bits "`11`", the first index (`0`) is the "Entity" type, and the second index (`1`) would be the "Parent" type. When an entity applies it's effective state, it will become effectively active if both bits are set active: "`11`", or effectively inactive if any one or both bits are set inactive: "`01`", "`10`", "`00`".
```
ShaunaGordon, line 57: Might want to include a matrix to make it easier to skim the page and reference what combination produces which state.

ShaunaGordon (changes requested):
A little bit of formatting and phrasing feedback, but otherwise, looks good.

Be sure to update this branch with the current upstream development branch, just to make sure it doesn't cause problems.

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: Your first assumption is the correct. Every entity everywhere has a bitmask.
The entity is only activated when ALL the bits are set to active, as any single bit requiring inactive would drive it to be inactive. The core purpose being that setting any flag preserves the state of others, so there's no destruction of state.

Say you have an entity, It's local state is active, it's parent state is active, but it's part of a streaming system. The streamer says: "Stream flag is inactive". That makes it so the entity is inactive. When it gets into frame and the stream says: "Streaming flag is okay now." The entity returns to it's state, unchanged.

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: All 1's produces active. Any 0 becomes inactive.

content/docs/user-guide/programming/components/entity-activation-system.md
ShaunaGordon: Is the bitmask a subset of the 32bit allocation? Or do each get a 32bit thing for other things to hook into?

Is there a way to illustrate this with a visual of some sort, perhaps?

content/docs/user-guide/programming/components/entity-activation-system.md
ShaunaGordon: It's mostly for skimming purposes (as opposed to clarification purposes). Right now, that info is a bit buried in a paragraph, but it seems like something people would want something they could glance at and get the information they need. If you think a different way to make this stand out for skimming would work better, that's fine, too.

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: The bitmask is defined in the system component through registrations that contextualize a bit.
Then when entities are operating, the various flags apply to whatever bit they were designated to.

image

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: How to visualize the Effective Active State.

image

I like to think of it as a lockpicking minigame.

ShaunaGordon: I really like these images, can you add them to the page itself, please?

GaianHelmers: On it! 👍

content/docs/user-guide/programming/components/entity-activation-system.md
GaianHelmers: Attempted to reword this. Makes more sense once you read about the finer details below.

content/docs/user-guide/programming/components/entity-activation-system.md
```diff
@@ -0,0 +1,75 @@
...
+
+![Visualization of how the entity determines it's active state.](/images/user-guide/programming/components/effective-active-per-instance.png)
```
ShaunaGordon:
```suggestion
![Visualization of how the entity determines its active state.](/images/user-guide/programming/components/effective-active-per-instance.png)
```

GaianHelmers: Got it, changed a header with the same mistake as well.

ShaunaGordon (approved):

PR #2747, opened 2025-11-12
Added small section about System Component Dependencies 

GaianHelmers:
## Change summary

Added small section about **System Component Dependencies** at the bottom of the System Component page.

Vital information to know when using System Components.

### Submission Checklist:


JT-SCB-GameDesign:
Excellent.  I expect we will be changing up the components information in multiple places for the Spring '26 main release.  So the location of this may change.  After a full review, we may want to put this in with other similar changes to cut down on noise.

So when were you thinking it would work best to get this into the docs?

GaianHelmers:
> So when were you thinking it would work best to get this into the docs?

It's immediately relevant. This is something many users are unaware of, and becomes a critical detail to setting up Render Passes. So the lack of access around this info is driving any users entering rendering work to fail.

I would want this topic shared in even more places quite frankly. But it's appropriate for System Components as it's most significant to the system component activation stack.

JT-SCB-GameDesign (approved):
Looks good as long as it's technically correct in expected use cases.
Preview page looks good.

PR #2782, opened 2026-05-26
Update dropdown

ShaunaGordon:
Update the version selector dropdown.


JT-SCB-GameDesign (approved):
Looks simple enough to me

PR #2789, opened 2026-05-27
Update modules for new api generation

ShaunaGordon:
Update references to the api module, so it's running the newest version.


JT-SCB-GameDesign (approved):
kk

